PR #4, opened 2019-02-10
Support for bayesian models

DominiqueMakowski:


DominiqueMakowski:
Seems to work for bayesian models. However it heavily relies on get_ and find_parameters 😀

PR #18, opened 2019-03-03
lme4 support

DominiqueMakowski:


DominiqueMakowski:
@strengejacke I took the liberty of re-using some of the sjstats code for p values 😬 Is that ok?

DominiqueMakowski:
Should the function be named `p_value()` or `p_values()` 🤔 ?

strengejacke:
Yes, feel free to use that code, I was planning to extract more parts from sjstats and put them into different easystats-packages, anyway. :-)

I'll lokk into this PR, some minor changes might be applies, e.g. for the argument-evaluation.

R/model_parameters.merMod.R
```diff
@@ -43,13 +44,17 @@ model_parameters.merMod <- function(model, ci = .95, standardize = FALSE, bootst
 
   # p value
   if("Pr(>|z|)" %in% names(parameters)){
-    names(parameters) <- gsub("Pr(>|z|)", "p", names(parameters))
+    names(parameters)[grepl("Pr(>|z|)", names(parameters))] <- "p"
```
strengejacke: Does this work without using `fixed = TRUE`? I think you else need to escape special regex chars.

DominiqueMakowski: Yes, I would have to escape the characters for it to work, but I found it was quicker to use this alternative solution instead of struggling with the escape codes... 😅 
It is not less efficient, is it?

strengejacke: I think it would be more safe to use `[grepl("Pr(>|z|)", names(parameters), fixed = TRUE)`, because `(` and `|` are special regex chars in your pattern-argument.

DominiqueMakowski: You mean like this : `names(parameters)[grepl("Pr(>|z|)", names(parameters), fixed = TRUE)] <- "p"`?

strengejacke: Yes!

DominiqueMakowski: done :)

DominiqueMakowski:
I will merge this initial implementation we can still improve it later

PR #20, opened 2019-03-07
Distributions-related functions and ANOVAs support

DominiqueMakowski:
This PR adds a few distribution-related functions (motivated by necessities at work):
- kurtosis and skewness
- smoothness: methods for quantifying smoothness of a vector
- normalize: variant of standardize
- `find_distribution()`: this function is based on an ML model (which will be improved as soon as I'll have access to servers) that classifies the distribution.


DominiqueMakowski:
@strengejacke I would tend to think that eta squared and others would belong here, as they are more parameters-related metrics than general model performance indices.

As from what I understood, the guidelines for reporting anovas suggest the use of omega squared over eta squared, I thought I would start with this one, as it is the one that would be used by default by report. Nevertheless, in the future, yes, when we have the time it would be good to add other metrics as well, since performance and parameters are meant to be rather "guidelines agnostic" (providing all tools and not forcing too much the usage of the good ones).

strengejacke:
I'm just asking because then I could remove those functions from _sjstats_, if they completely mimic the behaviour and features, so I can reduce sjstats and focus on specific tasks... As you said, sjstats does alot, and nobody even knows what...

For confidence intervals of omega squared, you may also need bootstrapping. You may look at this code:
https://github.com/strengejacke/sjstats/blob/master/R/eta_sq.R

See docs here:
https://strengejacke.github.io/sjstats/reference/eta_sq.html

For eta-sq and partial omega-sq, the CI need to be bootstrapped.

strengejacke:
Oh, and everything that does not need to go into _performance_ is welcomed from my side, so we can finish the work on performance asap. :-)

DominiqueMakowski:
Agreed!

PR #24, opened 2019-03-21
update

DominiqueMakowski:


DominiqueMakowski:
I am still not sure how to expand the method in the best possible way regarding documentation

strengejacke:
I can't get `equivalence_test()` working. I'd like to re-import from bayestestR, so we don't need to define the S3-method again, somehow it doesn't work.

I don't want:

```r
equivalence_test <- function(...) {
  UseMethod("equivalence_test")
}
```

since we have this in bayestestR already...

PR #38, opened 2019-05-19
Update

DominiqueMakowski:


tests/testthat/test-model_parameters.BFBayesFactor.R
```diff
@@ -2,6 +2,8 @@ context("model_parameters.BFBayesFactor")
 
 test_that("model_parameters.BFBayesFactor", {
   library(BayesFactor)
+  library(dplyr)
+
   model <- BayesFactor::ttestBF(iris$Sepal.Length, iris$Sepal.Width, paired = TRUE)
   testthat::expect_equal(model_parameters(model)$BF, 6.107436e+69, tolerance = 2)
   model <- BayesFactor::ttestBF(formula = mpg ~ am, data = dplyr::mutate(mtcars, am = as.factor(am)))
```
strengejacke: Though it is only suggested, but if this is the only place where we need dplyr, can't we simply convert to factor before, and remove dplyr here?

DominiqueMakowski: We definitely can, it was a lazy move on my side where CRAN checks said "add dplyr to deps" so I added without further thought :)

PR #45, opened 2019-05-29
dev

DominiqueMakowski:


R/WIP_parameters_selection.R
```diff
@@ -0,0 +1,64 @@
...
+  formula
+}
+
+
+
+
+
+
+#' @keywords internal
+.reconstruct_formula <- function(parameters){
```
DominiqueMakowski: @strengejacke I wonder if there is a way of directly getting the "formula" instead of the parameters. If not, I wonder an exported function `construct_formula` or something (in insight?) could be useful to reconstruct an original and compact formula based on the parameters...

R/parameters_selection.R
```diff
@@ -0,0 +1,141 @@
...
+#' @rdname parameters_selection
+#' @export
+parameters_selection.merMod <- function(model, ...){
+  if (!requireNamespace("cAIC4", quietly = TRUE)) {
+    stop("Package `cAIC4` required. Please install it by running `install.packages(cAIC4)`.", call. = FALSE)
+  }
+
+  best <- cAIC4::stepcAIC(model)
+  # TODO: Need to improve reconstruct formula to also account for random effects and all
+  parameters <- insight::find_parameters(best$finalModel, effects = "fixed", flatten = TRUE)
```
strengejacke: Is here something we need to address in insight?

R/parameters_selection.R
```diff
@@ -0,0 +1,141 @@
...
+
+
+#' @rdname parameters_selection
+#' @export
+parameters_selection.merMod <- function(model, ...){
+  if (!requireNamespace("cAIC4", quietly = TRUE)) {
+    stop("Package `cAIC4` required. Please install it by running `install.packages(cAIC4)`.", call. = FALSE)
+  }
+
+  best <- cAIC4::stepcAIC(model)
```
strengejacke: This could go into performance.

R/parameters_selection.R
```diff
@@ -0,0 +1,141 @@
...
+
+  # Clean
+  if(tail(parameters, 1) == "sigma"){
+    parameters <- parameters[1:length(parameters)-1]
+  }
+  if(parameters[1] == "(Intercept)"){
+    parameters <- parameters[2:length(parameters)]
+  }
+
+  # Detect interactions
```
strengejacke: Insight now has a `find_interactions()` function.

R/WIP_parameters_selection.R
strengejacke: Do you mean find_formula()?

DominiqueMakowski: No really, if I'm not correct find_formula retrieve a formula from models and/or from terms, whereas here we would need to reconstruct the formula from the list of parameters. For instance from this:

```r
[1] "(Intercept)" "cyl"         "disp"        "cyl:disp"    "hp"          "sigma"    
```

to `cyl * disp + hp` 

Which gets tricky with triple or partial interactions

R/parameters_selection.R
DominiqueMakowski: This depends on if/what we want to do with the formula reconstruction

R/parameters_selection.R
DominiqueMakowski: Since it's really about parameters selection (rather than assessing or comparing AICs per se)  I feel like it fits here

R/parameters_selection.R
DominiqueMakowski: Although here we are working with a vector of parameters rather than a model

R/parameters_selection.R
strengejacke: Ok, but maybe we could think of adding the cAIC4 to performance?

R/parameters_selection.R
strengejacke: Not sure in which way you need to reconstruct a formula, I probably need to look a bit closer at the code.

R/parameters_selection.R
strengejacke: Ok.

R/parameters_selection.R
DominiqueMakowski: Sure, although I don't know what it's worth outside of variable selection, I haven't read the paper very thoroughly 😅

PR #52, opened 2019-06-18
Dev

DominiqueMakowski:


DominiqueMakowski:
Work pressure made me add `check_factorstructure` to run the two related checks, `check_kmo` (at first I wrote `check_KMO` but then I remembered the conventions :) and `check_sphericity`. These are used for factor analysis. 

Although it's not directly related to parameters, it has its place here alongside the soon-to-be-added dimensions reduction functions.

R/check_factorstructure.R
```diff
@@ -0,0 +1,127 @@
...
+#'
+#' @inheritParams check_kmo
+#' @examples
+#' library(parameters)
+#'
+#' check_factorstructure(mtcars)
+#'
+#' @seealso check_kmo check_sphericity
+#' @export
+check_factorstructure <- function(x, ...){
```
DominiqueMakowski: @strengejacke however, this way of simply running the two gives an ugly output (as they are not separated). I tried capturing the output, but then the colour disappears... any tips to address that?

DominiqueMakowski:
> Also, if you have any better idea for the name check_factorstructure. I thought of check_factorability (ugly), check_structure (too generic)

Actually, I am about to start re-implementing the `n_factors` function from psycho. Two alternatives:

- rename it as `check_factor_number` and then maybe also change `check_factor_structure` for continuity
- keep `n_factors` and create a new cluster of `n_*` functions returning the number. we already have the `n_obs` in insight. I am not sure for what else it could be used *yet* but we never know (now that I think of it `n_parameters` would actually be useful)

Thoughts?

R/check_factorstructure.R
strengejacke: Give me a short update, what do you want to achieve, and where's the current problem?

DominiqueMakowski: - I am slowly adding a new set of features in parameters around dimensionality reduction (including support for PCA / FA etc.)
- In this context, I added the two popular tests for checking if the data is suitable for FA
- I would like to have a master function that returns the output of these two checks "at once"
- A function simply running these two checks gives an ugly collated output
- how to improve it?

strengejacke: Ok, I'll look into this later today.

DominiqueMakowski:
Started rework on `principal_components`. TODO:

- [ ] Make `n` argument to work
- [ ] Add sorting capability (sort the row order based on their loadings
- [ ] Add method to automatically get the most representative component for each variable
- [ ] Improve print method (@strengejacke maybe we should think about adding a `print_table` method in insight to easily return nicely formatted tables (with separation lines), such as the ones in `report`?)

PR #75, opened 2019-07-21
final tweaks before CRAN submission

### LINKED ISSUE #74: degrees of freedom: DoF or df?
I've seen a trend (also in other languages such as Julia) toward naming degrees of freedom `dof` instead of `df` (my guess is that it's to avoid confusion with `dataframe`). However, in APA tables it's usually reported as `df`. 

Should we rename the column to `df`?

DominiqueMakowski:
close #74


DominiqueMakowski:
I'll submit this version today 💪

PR #77, opened 2019-07-25
dev

DominiqueMakowski:


DominiqueMakowski:
CRAN's new curator is quite comprehensive!
 
```
"The ownership of copyright and intellectual property rights of all 
components of the package must be clear and unambiguous (including from 
the authors specification in the DESCRIPTION file). Where code is copied 
(or derived) from the work of others (including from R itself), care 
must be taken that any copyright/license statements are preserved and 
authorship is not misrepresented.
Preferably, an ‘Authors@R’ would be used with ‘ctb’ roles for the 
authors of such code. Alternatively, the ‘Author’ field should list 
these authors as contributors.
Where copyrights are held by an entity other than the package authors, 
this should preferably be indicated via ‘cph’ roles in the ‘Authors@R’ 
field, or using a ‘Copyright’ field (if necessary referring to an 
inst/COPYRIGHTS file)."
f.i.: William Revelle

Please add references describing the methods in your package also to the 
description field of your DESCRIPTION file in the form
authors (year) <doi:...>
authors (year) <arXiv:...>
authors (year, ISBN:...)
with no space after 'doi:', 'arXiv:' and angle brackets for auto-linking.

Please add \value to .Rd files and explain the functions results in the 
documentation.

Some code lines in examples are commented out.
Please never do that. Ideally find toy examples that can be regularly 
executed and checked. Lengthy examples (> 5 sec), can be wrapped in 
\donttest{}.

\dontrun{} should be only used if the example really cannot be executed 
(e.g. because of missing additional software, missing API keys, ...) by 
the user. That's why wrapping examples in \dontrun{} adds the comment 
("# Not run:") as a warning for the user.
Please unwrap the examples if they are executable in < 5 sec, or create 
additionally small toy examples to allow automatic testing.
As you have implemented tests already you could also replace \dontrun{} 
with  \donttest{}.

Please fix and resubmit.
```

strengejacke:
You can take examples for the DESCRIPTION (authors, doi) from performance, where we had the same feedback: https://github.com/easystats/performance/blob/master/DESCRIPTION

PR #78, opened 2019-07-26
dev

DominiqueMakowski:
- [x] 3
```
Please add \value to .Rd files and explain the functions results in the 
documentation.
```

- [x] 4
```
Some code lines in examples are commented out.
Please never do that. Ideally find toy examples that can be regularly 
executed and checked. Lengthy examples (> 5 sec), can be wrapped in 
\donttest{}.

\dontrun{} should be only used if the example really cannot be executed 
(e.g. because of missing additional software, missing API keys, ...) by 
the user. That's why wrapping examples in \dontrun{} adds the comment 
("# Not run:") as a warning for the user.
Please unwrap the examples if they are executable in < 5 sec, or create 
additionally small toy examples to allow automatic testing.
As you have implemented tests already you could also replace \dontrun{} 
with  \donttest{}.

Please fix and resubmit.
```


DominiqueMakowski:
Will re-submit to CRAN

DominiqueMakowski:
Erf, cannot submit to CRAN because it uses `reshape_ci` which is not on bayestestR's CRAN version 😕 😕

DominiqueMakowski:
In this case, let's submit `parameters` right after bayestestR and `see`, hopefully in a week, so we improve and demonstrate some plotting abilities

tests/testthat/test-model_parameters.mixed.R
```diff
@@ -6,11 +6,14 @@ test_that("model_parameters.mixed", {
...
+
   params <- model_parameters(insight::download_model("merMod_1"), standardize = "refit")
   testthat::expect_equal(c(nrow(params), ncol(params)), c(2, 8))
 
   params <- model_parameters(insight::download_model("merMod_2"), standardize = "refit")
   testthat::expect_equal(c(nrow(params), ncol(params)), c(3, 8))
 
-  # TODO: Not sure how to deal with bootstrapped mixed models... As it throws a reasonable amount of singular fits...
+  # TODO: Not sure how to deal with bootstrapped mixed models... As it throws an unreasonable amount of singular fits...
 })
```
strengejacke: The mtcars data set is probably no good choice for testing in general, because it has no real "binary" variables to test logistic regression, nor enough cases for mixed models. If you bootstrap, you even get duplicates from these few cases, which causes all the singular fits.

We should think of also using other data sets...

PR #88, opened 2019-08-10
dev

DominiqueMakowski:


DominiqueMakowski:
https://cran.r-project.org/submit.html

Unfortunately, it got blocked until the 18th -_- -_- -_- -_-

strengejacke:
> Unfortunately, it got blocked until the 18th

Why that? Too many tries? Haven't you done your homework before submission? ;-)

strengejacke:
If I understood right, one request was to add people to the authors from which we have taken sourcecode, no?

DominiqueMakowski:
> one request was to add people to the authors from which we have taken sourcecode, no?

Yes, although I think we can try explaining that these people are mentioned to credit them and for traceability purposes (we could simply imply import the packages as dependencies but in order to keep the latter to a minimum, we simply included and re-adapted some specific function) but we're not really comfortable including people as authors without asking them and without their explicit contribution to that package

PR #111, opened 2019-09-15
dev

DominiqueMakowski:


R/utils_pca_efa.R
```diff
@@ -1,62 +1,73 @@
...
 #' @export
 summary.parameters_efa <- function(object, ...) {
-  insight::print_color("# (Explained) Variance of Principal Components\n\n", "blue")
 
   x <- attributes(object)$summary
-  x$Variance_Explained <- x$Variance / sum(x$Variance)
-  x$Std_Dev <- attributes(object)$model$sdev
-
-  col_names <- x$Component
+  # x$Std_Dev <- attributes(object)$model$sdev
```
DominiqueMakowski: @strengejacke what is this for? Is it actually used somewhere? Couldn't find any instances

strengejacke: Not sure, that must be your code... 😕

DominiqueMakowski: don't think so, would have never call it "Std_Dev" :grin:

PR #137, opened 2019-10-04
dev

DominiqueMakowski:
- The model parameters table also prints a blue title if object has a title attribute
- This title can be automatically found using the format_model and format_algorithm functions that have been copied from report to here


DominiqueMakowski:
Merging this, next PR will start reorganizing stuff

Issue #138, opened 2019-10-07
Effect size package?


strengejacke:
We have quite some stuff dealing with effect sizes, conversion and ANOVA stuff. What do you think, does it make sense to separate these functions from _parameters_ and put them into an own package?

DominiqueMakowski:
IMO effect sizes for models are closely related to parameters (being just scaleless parameters), so, for now, I'd say we can keep the handful of conversion functions here (as they are quite straightforward to maintain) and are a core aspect of the parameters engineering scope of the package. That being said, if we start to extend this aspect (for instance metaanalysis tools etc.) then that might become necessary indeed...

(but let's get the 3 remaining packages on CRAN before adding more 😅)

DominiqueMakowski:
Mmh after spending my morning overthinking standardized parameters and all, I think you might be right. We could create a separate package containing methods for standardizing, converting and interpreting parameters (mainly move functions from here and from report). That would ease their subsequent inclusion in parameters and report. It would also be easier to document it there.

I'll create the new package, and copy the relevant functions *without* removing them from parameters and report until the package in on CRAN. We'll see how it works having this package separate.

This could go fast, as most functions work already. @strengejacke are you okay to take over parameters and be the one listed as maintainer?

strengejacke:
I would probably leave standardization in parameters? Making effectsize a similar small package like correlation?

DominiqueMakowski:
standardization is needed for effet sizes, but not needed for parameters. And it better fits there IMO. But as `effectsize` would be super light (no dependencies (not even suggested) aside from insight and (?) bayestestR), it's not a problem to import it in the others

strengejacke:
True.

strengejacke:
Although standardization is something I often use in modelling, while I actually never use effect size (conversion).

Issue #110, opened 2019-09-14
Improvements to Effect Size (Eta2, Omega2) Indices for ANOVAs


DominiqueMakowski:
Combining several issues here:

- [ ] Support repeated measures ANOVAs (#84, #100):
    - make it work for `anova(themixedmodel)`
    - Make sure these are *correct*
- [ ] Generalized eta2 and Omega2 (#91)
    - https://www.frontiersin.org/articles/10.3389/fpsyg.2013.00863/full
- [ ] Adjusted eta2
    - https://journals.sagepub.com/doi/abs/10.1177/2515245919855053?journalCode=ampa

Does anybody know an expert of ANOVAs that could help us going through this?

IndrajeetPatil:
@DominiqueMakowski Maybe these functions can also be moved to the new `effectsize` package?

DominiqueMakowski:
I don't see why not!

mattansb:
I agree - `effectsize` seems like the perfect place for all `r`, `d`, `OR`, and `vaf` measures!

DominiqueMakowski:
I'll need help 🙉

IndrajeetPatil:
Until these generalized functions are implemented in `effectsize` itself, we can probably just move (literally copy-paste) the eta- and partial-omega squared functions to `effectsize` and re-export them in `parameters`? 
(Of course, this means that `effectsize` will have to go on `CRAN` soon, so this can wait if there is a new version of `parameters` in the offing.)

All the new development for these new generalized functions can then happen in `effectsize` itself.

DominiqueMakowski:
Yup, I just wonder whether the eta2 & co functions need stuff from parameters that would need to be transferred too. 

For instance in parameters a line added the SE to the object:

https://github.com/easystats/effectsize/blob/28fa79ca7486b51b33aa4ce9c72992e1a3a95a18/R/standardize_parameters.R#L145

But I had to deactivate it because it reaquires `standard_error`, which is parameters... Not sure how to proceed in such cases. Do we move `standard_error` as well? But standard_error relies in some cases on the whole model_parameters. So we need to find a better way...

mattansb:
Perhaps under `effectsize` some of these functions would be stripped down a bit (e.g., `.standardize_parameters_posthoc` wouldn't return standardized SEs?).

Additionally, some functions can appear, maybe internally / not-exported, in more than 1 package, if need be (obviously this should be avoided as much as possible).

DominiqueMakowski:
[...]

PR #139, opened 2019-10-09
dev - Reorganization of parameters (effectsize)

DominiqueMakowski:
This PR will track the reorganization of parameters related to what has been decided #110 #138 

***~Do not merge this until effectsize is on CRAN~***

The goal is to transfer things related to effect sizes there.


DominiqueMakowski:
The question is, what is the best order to update on CRAN? parameters -> effectsize -> see?

DominiqueMakowski:
The ANOVA effect sizes functions are based on `.extract_parameters_anova()` , which is currently an internal. Should we export the `extract_parameters_*` functions (but document them as being mostly dev-related and that user should consider using `model_parameters` instead)?

strengejacke:
> Do not merge this until effectsize is on CRAN

pkg effectsize needs `dof()`, so we _have_ to merge this PR _before_ effectsize is on CRAN, no? But this won't work, as it breaks reverse dependencies.

DominiqueMakowski:
In theory yes, it should be merged after effectsize contrary to what the message said.

One of the possibilities I thought about is to submit effectsize just before parameters, with all the tests and examples that require parameters deactivated. This way, it would pass on CRAN before parameters is updated, so we can then quickly submit updated parameters and sjpacks. What do you think?

strengejacke:
Ok, sounds good. I trust you, because I couldn't follow the development of effectsize in detail last week ;-)

DominiqueMakowski:
> I trust you

\**starts to sweat*\* 😅 should be alright hopefully :)

Let's start by submitting insight and bayetestR

strengejacke:
I'm still not 100% convinced that standardize and normalize should also go into effectsize, especially since effectsize can easily import or re-export those functions from parameters.

and, model_parameters() is losing its feature to show std. coefficients...

mattansb:
> and, model_parameters() is losing its feature to show std. coefficients...

😨😨😨😨😨 I thought `parameters` was dep on `effectsize`??

Also, why are the `interpret_*` functions in `effectsize`? Shouldn't they be in `report`??

mattansb:
> and, model_parameters() is losing its feature to show std. coefficients...

I would consider this a breaking change - it will definitely break my code :/

I know we can't have circular dependency - but can't parameters suggest `effectsize`, and have the whole `if(!require("effectsize"))` thing?

strengejacke:
Yeah, but it's not planned for the forthcoming update of parameters and initial release of effectsize.

strengejacke:
My feeling is that standardization is more a part of the parameters package than effectsize. _effectsize_ comprises all the anova-effect-size and effect-size-conversion stuff, which is not needed for model parameters. however, standardization can be useful for both packages, so it should live in the "more basic" package. If standardization is in parameters, it can be used by parameters _and_ effectsize. Else, it can only be used in effectsize. Don't you think so, too?

DominiqueMakowski:
Should we remove all the ANOVA ES from here as well to avoid conflicts once effectsize will be up?

strengejacke:
Yes, I think we can remove all effectsize stuff, even for the next update. It shouldn't conflict with any reverse dependencies.

strengejacke:
Based on https://github.com/easystats/parameters/issues/142#issuecomment-544124870, we can merge this now, and I will then "re-implement" the standardized options based on suggesting effectsize, right?

DominiqueMakowski:
I'd say, altough we might want to remove the ANOVA ES things before merging and try to have that travis passing just to make sure that the new version "works"

strengejacke:
What are the "ANOVA ES" things?

DominiqueMakowski:
ANOVA effect sizes functions that have been copied to effectsize but not removed from here

strengejacke:
Ah, I see. Thought these have already been removed as well. Yes yes, away with them ;-)

strengejacke:
> I'd say, altough we might want to remove the ANOVA ES

That was actually my main intentsion when I suggested the effect size package :-D

PR #146, opened 2019-10-28
Dev

DominiqueMakowski:


strengejacke:
Why have you removed `stats::` from predict?

DominiqueMakowski:
Because predict.kmeans doesn't exist, hence I added this method in parameters

strengejacke:
yes, but `predict.kmeans()` is an S3-method, and it's called via `stats::predict()` on kmeans-objects?

DominiqueMakowski:
True

PR #205, opened 2020-02-05
enhance n_clusters

strengejacke:


strengejacke:
Can you add the pkg to the DESCRIPTION as well?

DominiqueMakowski:
Not really, it's not on CRAN 😱

strengejacke:
Will that ever pass any check?

DominiqueMakowski:
Well yes because we don't use it in the EXAMPLE 😏

DominiqueMakowski:
we'll see, worst case scenario we remove it

strengejacke:
> The ‘Additional_repositories’ field is a comma-separated list of repository URLs where the packages named in the other fields may be found. It is currently used by R CMD check to check that the packages can be found, at least as source packages (which can be installed on any platform).

strengejacke:
We may add 

`Additional_repositories: https://bioconductor.org/packages/release/bioc/html/M3C.html`

to the DESCRIPTION file?

DominiqueMakowski:
I didn't know this field existed, and wonder how is it different from "Remotes". But it sounds appropriate indeed

strengejacke:
I'LL ask at r-package-devel (https://stat.ethz.ch/mailman/listinfo/r-package-devel), but currently, I always get a timeout when I try to subscribe...

strengejacke:
haha, that happens when R relies on Python... The list is

image

so no wonder it doesn't work properly ;-)

strengejacke:
Ok, I got following answer:

> BioC is a standard repository that you do not need to delcare.

So, we can/should be safe to add M3C to "Suggest".

DominiqueMakowski:
does it mean that it knows automatically that if it's not from CRAN it needs to be downloaded from bioconductor?

strengejacke:
> does it mean that it knows automatically that if it's not from CRAN it needs to be downloaded from bioconductor?

Yes, this is how I understood Uwe.

DominiqueMakowski:
Interesting

R/format_p.R
```diff
@@ -26,7 +26,7 @@ format_p <- function(p, stars = FALSE, stars_only = FALSE, name = "p", missing =
       ifelse(p < 0.001, "< .001***",
         ifelse(p < 0.01, "< .01**",
           ifelse(p < 0.05, "< .05*",
-            paste0("= ", insight::format_value(p, digits))
+            paste0("= ", insight::format_value(p, 3))
```
strengejacke: Why this change?

strengejacke: ok, I assume this is an APA requirement?

DominiqueMakowski: Otherwise it fails because digits is a character ("apa") and format_value expects an int

strengejacke:
Dear pull-requester,
see the errors related to your current PR. Please fix asap, latest before 8. Feb. 2020 to safely retain your PR on GitHub.
Sincerly,
easystats maintainance

DominiqueMakowski:
I knew it! @strengejacke was a bot this whole time 🤖

DominiqueMakowski:
I've been so used to seeing parameters, performance and insight fail travis that I didn't even think it could come from my commit 😬

strengejacke:
> I've been so used to seeing parameters, performance and insight fail travis

Wait, that must have been **years** ago. insight is fine, parameters as well, and performance usually, too. It currently fails because it needs the latest _parameters_ on CRAN (which I plan to submit next week, because I need to submit performance to CRAN).

strengejacke:
I think you need to merge master into dev, and then add M3C as suggested, and that's it.

DominiqueMakowski:
There was a missing closing quote in my commit as well 🙄  fixing now

strengejacke:
when this PR passes, I'll merge and include it in the next update, or not? can you also update the news, please? 😬

DominiqueMakowski:
>  I'll merge and include it in the next update

Yes, yes, necessary for report!

PR #307, opened 2020-09-26
set.seed in `data_partition()`

zen-juen:
It seems that the EFA factor loadings are different each time the data is split into training and test sets using `data_partition()`. We suspect this is because the sets are split differently with each run @DominiqueMakowski 

This should be easily resolved with `set.seed()` before running `sample()`

### DIFF
--- a/R/data_partition.R
+++ b/R/data_partition.R
@@ -56,6 +56,7 @@ data_partition <- function(x, training_proportion = 0.7, group = NULL) {
 
 #' @keywords internal
 .data_partition <- function(x, training_proportion = 0.8) {
+  set.seed(333)
   training_indices <- sample(1:nrow(x), size = training_proportion * nrow(x))
   test_indices <- (1:nrow(x))[-training_indices]
 
--- a/R/data_partition.R
+++ b/R/data_partition.R
@@ -16,7 +16,7 @@
 #' head(data_partition(df, group = "Species"))
 #' head(data_partition(df, group = c("Species", "Smell")))
 #' @export
-data_partition <- function(x, training_proportion = 0.7, group = NULL) {
+data_partition <- function(x, training_proportion = 0.7, group = NULL, seed = NULL) {
   if (!is.data.frame(x)) {
     x <- tryCatch(
       expr = {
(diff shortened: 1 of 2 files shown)


mattansb:
Hi!
Thanks for the PR!

It is generally bad practice to set the RNG seed for users, as this can affect things further down the user's workflow.
If we do go this route, we would have to also reset the seed to what it was before we set it to a specific value inside the function.

@DominiqueMakowski , thoughts?

strengejacke:
We could use an argument to set the seed?

mattansb:
Users could set the seed before running the function themselves, just as well, no?

strengejacke:
Usually, yes. I don't think it's a good idea to use a fixed seed for data partitioning. And users could set their own seed in advance. Sometime, having a `seed` argument makes it more "obvious" that results are "random" (like in rstanarm's or brms' modelling functions, where you have a seed argument, I think). So that might be the main purpose of such an argument here.

mattansb:
Yes, that sounds like a good idea (:

DominiqueMakowski:
Yes indeed, what I had in mind was having a seed argument defaulting to NULL that the user can specify. The problem of global seeds @mattansb is that sometimes they are not propagated reliably in for instance Rmardown docs (from chunks to chunks), so sometimes it's just easier to provide a seed to the function.

@zen-juen you can try modifying this PR and add a `seed=NULL` argument to the function definition, and then something like `if(is.null(seed)==FALSE){set.seed(seed)}`

zen-juen:
Ah yes, a seed argument is definitely more sensible to address this issue 😄 
Thanks for the suggestions! I've just updated the PR, let me know if there's anything else I should modify

IndrajeetPatil:
@zen-juen You also need to run `devtools::document()`, so that the documents are updated to reflect the changes to the function. This is the reason why all the builds are failing.

DominiqueMakowski:
the check fails seem unrelated to the changes if I'm not mistaken

IndrajeetPatil:
> the check fails seem unrelated to the changes if I'm not mistaken

Now, yes. Now the failures are related to `show_sigma` not currently being documented. 
So, if nothing else is needs to be changed, this can be merged.

strengejacke:
Thanks!

Issue #308, opened 2020-09-28
adding support for `pairwise.htest` objects


IndrajeetPatil:
@strengejacke What do you think about this? Does the function implementation make sense?

# Function

```r
# function
model_parameters.pairwise.htest <- function(model, ...) {
  m <- model$p.value
  result <-
    data.frame(
      group1 = rep(rownames(m), each = ncol(m)),
      group2 = rep(colnames(m), times = nrow(m)),
      p.value = as.numeric(base::t(m)),
      p.adjust.method = model$p.adjust.method,
      stringsAsFactors = FALSE
    )
  stats::na.omit(result)
}
```

# Examples

```r
# t-test
attach(airquality)
Month <- factor(Month, labels = month.abb[5:9])
x <- pairwise.t.test(Ozone, Month)

model_parameters.pairwise.htest(x)

#>    group1 group2      p.value p.adjust.method
#> 1     Jun    May 1.0000000000            holm
#> 5     Jul    May 0.0002638036            holm
#> 6     Jul    Jun 0.0511274135            holm
#> 9     Aug    May 0.0001949061            holm
#> 10    Aug    Jun 0.0498733313            holm
#> 11    Aug    Jul 1.0000000000            holm
#> 13    Sep    May 1.0000000000            holm
#> 14    Sep    Jun 1.0000000000            holm
#> 15    Sep    Jul 0.0048787981            holm
#> 16    Sep    Aug 0.0038781084            holm

# prop test
smokers  <- c( 83, 90, 129, 70 )
patients <- c( 86, 93, 136, 82 )
p <- pairwise.prop.test(smokers, patients)

model_parameters.pairwise.htest(p)
#>   group1 group2    p.value p.adjust.method
#> 1      2      1 1.00000000            holm
#> 4      3      1 1.00000000            holm
#> 5      3      2 1.00000000            holm
#> 7      4      1 0.11856482            holm
#> 8      4      2 0.09321728            holm
#> 9      4      3 0.12376805            holm

# wilcox test
attach(airquality)
Month <- factor(Month, labels = month.abb[5:9])
## These give warnings because of ties :
w <- pairwise.wilcox.test(Ozone, Month, p.adj = "bonf")

model_parameters.pairwise.htest(w)
#>    group1 group2     p.value p.adjust.method
#> 1     Jun    May 1.000000000      bonferroni
#> 5     Jul    May 0.000299639      bonferroni
#> 6     Jul    Jun 0.141362469      bonferroni
[...]

PR #324, opened 2020-11-05
close #251

strengejacke:
close #251

PR #327, opened 2020-11-08
support pairwise h test #308; use testthat 3rd ed

IndrajeetPatil:
- adds initial support for pairwise h-test objects
- tests for the same
- `NEWS.md` modified and adopts 80-characters limit
- adopts testthat 3.0 conventions
- retains only stricter tests for aov-ci functionality (since this tests all aspects of the output)


tests/testthat/test-betareg.R
```diff
@@ -1,9 +1,13 @@
 if (Sys.getenv("USER") != "travis") {
   if (require("testthat") && require("parameters") && require("betareg")) {
+    library(betareg)
```
strengejacke, line 3: we don't need the library() call here

tests/testthat/test-betareg.R
```diff
@@ -37,12 +41,20 @@ if (Sys.getenv("USER") != "travis") {
     test_that("p_value", {
       expect_equal(
         p_value(m1)$p,
-        c(0, 0, 0, 0, 0, 0, 0, 0, 1e-05, 0.00114, 0, 6e-05),
+        c(
+          3.44252055464711e-250, 2.59222187511485e-65, 3.33610354513962e-29,
```
strengejacke, line 45: why did you add this precision?

tests/testthat/test-model_parameters.aov_es_ci.R
```diff
@@ -1,614 +1,436 @@
-if (require("insight") && require("effectsize") && require("testthat") && require("lme4") && require("parameters")) {
```
strengejacke, line 1: why have these been removed?

strengejacke (changes requested):

tests/testthat/test-model_parameters.aov_es_ci.R
IndrajeetPatil: I am not sure what these tests add anything over and above the stricter tests. The stricter tests covers all aspects of the returned output.

tests/testthat/test-betareg.R
IndrajeetPatil: Because the tests were failing with `testthat 3.0.0` and doing this fixed it (at least on my Windows machine).

strengejacke: Ok, got it. But maybe you should not skip them everywhere? It should at least run on GitHub-CI.

strengejacke (changes requested):

tests/testthat/test-model_parameters.aov_es_ci.R
strengejacke: Ok, got it. But maybe you should not skip them everywhere? It should at least run on GitHub-CI.

strengejacke:
And there's a generated PDF in the testthat folder that should be removed. Was it from local tests?

IndrajeetPatil:
Why should it be removed? 
Running `devtools::test()` locally is always going to generate that file, so we might as well make peace with it. 
Or remove the tests that generate graphics.

strengejacke:
Not sure if we need to add this file to rbuildignore? I personally don't like these temporary files ;-)

IndrajeetPatil:
Yeah, me neither. But it's a pain to keep remembering to delete the file each time. Just ignoring it is better.

IndrajeetPatil:
Is there a way to skip a certain test only on Ubuntu?

strengejacke:
> Yeah, me neither. But it's a pain to keep remembering to delete the file each time. Just ignoring it is better.

Better to put them on gitignore, actually. Just came to my mind.

strengejacke:
> Is there a way to skip a certain test only on Ubuntu?

```r
osx <- tryCatch({
  si <- Sys.info()
  if (!is.null(si["sysname"])) {
    si["sysname"] == "Darwin" || grepl("^darwin", R.version$os)
  } else {
    FALSE
  }
})
```

replace with Ubuntu-OS.

strengejacke:
Maybe you should open a separate PR kust for the support of pairwise.htest, so that can be merged. Not sure if testthat 3rd ed. is stable enough for the CI?

IndrajeetPatil:
But this failure actually has nothing to do with switching to `testthat 3.0`. 

This exact same thing was also happening [before](https://github.com/easystats/parameters/pull/324) and that is precisely why we were skipping these tests so far. 
So we must address this now.

IndrajeetPatil:
Also, the test failure here: https://github.com/easystats/parameters/pull/327/checks?check_run_id=1370844389
has nothing to do with any of the changes I made. I am not sure what's going on there.

IndrajeetPatil:
I think this can be merged, and we can conditionally skip the aov-ci related tests later. 
They don't fail either on `MacOS` or on `Windows`, so we can be confident that we are not breaking anything.

IndrajeetPatil:
I saw you addressed this in a different PR. Should I just close this one then?

strengejacke:
Yes, I copied your pairwise.htest-stuff to have this working, and we can open a new PR for the transition to testthat ed. 3. We should try to make all tests pass, as it is know.

PR #342, opened 2020-11-16
closes #337

### LINKED ISSUE #337: supporting `verbose` argument for all methods
For some models this argument is unsupported, which can result in an error. I think we should add this argument regardless of whether it is actually relevant for the method or not so that it is not passed to the regression function. 

``` r
library(lqmm)

set.seed(123)

M <- 50
n <- 10
test <- data.frame(x = runif(n*M,0,1), group = rep(1:M,each=n))
test$y <- 10*test$x + rep(rnorm(M, 0, 2), each = n) + rchisq(n*M, 3)
fit.lqmm <- lqmm(fixed = y ~ x, random = ~ 1, group = group,    
                 data = test, tau = 0.5, nK = 11, type = "normal")

parameters::model_parameters(fit.lqmm) 
#> Parameter   | Coefficient |   SE |        95% CI | t(49) |      p
#> -----------------------------------------------------------------
#> (Intercept) |        3.44 | 0.59 | [2.25,  4.64] |  5.79 | < .001
#> x           |        9.26 | 0.55 | [8.16, 10.36] | 16.95 | < .001

parameters::model_parameters(fit.lqmm, verbose = FALSE) 
#> Error in boot.lqmm(object, ...): unused argument (verbose = FALSE)
```

For instances, for a function like following, the function will work for some objects, while not for others because of `verbose` not being present for some methods:

```r
tidy_model_parameters <- function(model, ...) {
  parameters::model_parameters(model, verbose = FALSE, ...) %>%
    parameters::standardize_names(data = ., style = "broom") %>%
    as_tibble(.)
}
```

IndrajeetPatil:


IndrajeetPatil:
Is this what you had in mind @strengejacke? If yes, I will do the same of the rest of the files.
Just wanted to double-check before proceeding ahead.

strengejacke (approved):

strengejacke:
Yes, that was what I was thinking of. It's just to not break functionality, if `verbose` is passed to `model_parameters()` and passed down via `...` when there is no `verbose` argument. We can add functionality, if necessary and where applicable, bit by bit.

IndrajeetPatil:
> Error: Vignette re-building failed.

Okay, the build failures have nothing to do with the changes I made, so I am going to go ahead and merge this.

PR #352, opened 2020-11-27
further steps to increase code readability

IndrajeetPatil:


R/factor_analysis.R
```diff
@@ -19,27 +20,40 @@
 #'   efa <- factor_analysis(mtcars[, 1:5], n = 2)
 #'   summary(efa)
 #'   predict(efa)
-#' \donttest{
```
strengejacke, line 22: Why did the tool change the examples here? This doesn't increase credibility of that tool. ;-)

R/cluster_analysis.R
```diff
@@ -60,9 +60,28 @@
 #' }
 #' @importFrom stats dist na.omit hclust kmeans cutree complete.cases
 #' @export
-cluster_analysis <- function(x, n_clusters = NULL, method = c("hclust", "kmeans"),
-                             distance = c("euclidean", "maximum", "manhattan", "canberra", "binary", "minkowski"),
-                             agglomeration = c("ward", "ward.D", "ward.D2", "single", "complete", "average", "mcquitty", "median", "centroid"),
+cluster_analysis <- function(x,
+                             n_clusters = NULL,
```
strengejacke, line 64: I would rather start the line earlier (at position 2 or so) instead of so many line-breaks, because too much vertical scrolling isn't so much better than horizontally scrolling, imho.

R/model_parameters.psych.R
```diff
@@ -27,11 +28,11 @@
 #'
 #'   principal_components(attitude, n = 3, sort = TRUE, threshold = 0.2)
 #'
-#' \donttest{
+#'
 #'   # Exploratory Factor Analysis (EFA) ---------
 #'   efa <- psych::fa(attitude, nfactors = 3)
 #'   model_parameters(efa, threshold = "max", sort = TRUE, labels = as.character(1:ncol(attitude)))
-#' }
```
strengejacke, line 34: Same here.

R/model_parameters.psych.R
```diff
@@ -41,16 +42,14 @@
 #' }
 #'
 #' # FactoMineR ---------
-#' \dontrun{
```
strengejacke, line 44: Are examples not run when they're wrapped in `donttest{}`? Else, this should be reverted.

R/n_factors.R
```diff
@@ -44,200 +44,219 @@
...
+      cor <- stats::cor(x, use = "pairwise.complete.obs")
+    }
+    eigen_values <- eigen(cor)$values
+
+
+    # Initialize dataframe
+    out <- data.frame()
+
+    # nFactors -------------------------------------------
+    if ("nFactors" %in% c(package)) {
```
strengejacke, line 74: We could remove the `c()` here.

R/bootstrap_parameters.R
```diff
@@ -46,7 +59,15 @@ bootstrap_parameters <- function(model, iterations = 1000, centrality = "median"
     p_value <- FALSE
   }
 
-  parameters <- bayestestR::describe_posterior(data, centrality = centrality, ci = ci, ci_method = ci_method, test = test, ...)
+  parameters <-
```
strengejacke, line 62: Not sure if we can add a style-rule here, but I find it odd to have a line-break after the assignment arrow. I would rather format into something like:

```r
parameters <- bayestestR::describe_posterior(
  data,
  centrality = centrality,
  ci = ci,
  ci_method = ci_method,
  test = test,
  ...
)
```

R/factor_analysis.R
```diff
@@ -74,13 +88,27 @@ factor_analysis.data.frame <- function(x, n = "auto", rotation = "none", sort =
 
   # Pass cor if available
   if (!is.null(cor)) {
-    out <- model_parameters(psych::fa(cor, nfactors = n, rotate = rotation, n.obs = nrow(x), ...), sort = sort, threshold = threshold)
-  } else{
-    out <- model_parameters(psych::fa(x, nfactors = n, rotate = rotation, ...), sort = sort, threshold = threshold)
+    out <-
```
strengejacke, line 91: See comment above

R/n_factors.R
```diff
@@ -427,121 +446,133 @@ print.n_clusters <- print.n_factors
...
 
 #' @keywords internal
-.n_factors_vss <- function(x = NULL, cor = NULL, nobs = NULL, type = "FA", rotation = "varimax", algorithm = "default") {
-  if (algorithm == "default") {
-    if (tolower(type) %in% c("fa", "factor", "efa")) {
-      algorithm <- "minres"
-    } else {
-      algorithm <- "pc"
+.n_factors_vss <-
+  function(x = NULL,
```
strengejacke, line 450: see above

R/n_factors.R
```diff
@@ -427,121 +446,133 @@ print.n_clusters <- print.n_factors
...
 
 
 #' @keywords internal
-.n_factors_fit <- function(x = NULL, cor = NULL, nobs = NULL, type = "FA", rotation = "varimax", algorithm = "default") {
-  if (algorithm == "default") {
-    if (tolower(type) %in% c("fa", "factor", "efa")) {
-      algorithm <- "minres"
-    } else {
-      algorithm <- "pc"
+.n_factors_fit <-
```
strengejacke, line 501: see above

R/n_factors.R
```diff
@@ -553,51 +584,69 @@ print.n_clusters <- print.n_factors
 
 #' @importFrom stats lm
 #' @keywords internal
-.nBentler <- function(x, N, model = model, log = TRUE, alpha = 0.05, cor = TRUE, details = TRUE, ...) {
-  if (!requireNamespace("nFactors", quietly = TRUE)) {
-    stop("Package 'nFactors' required for this function to work. Please install it by running `install.packages('lattice')`.")
-  }
+.nBentler <-
```
strengejacke, line 587: see above

R/n_parameters.R
```diff
@@ -139,15 +142,30 @@ n_parameters.vgam <- n_parameters.gam
 
 #' @rdname n_parameters
 #' @export
-n_parameters.brmsfit <- function(x, effects = c("all", "fixed", "random"), component = c("all", "conditional", "zi", "zero_inflated", "dispersion", "simplex", "sigma", "smooth_terms"), ...) {
+n_parameters.brmsfit <- function(x,
+                                 effects = c("all", "fixed", "random"),
+                                 component = c(
```
strengejacke, line 147: See comment above

R/principal_components.R
```diff
@@ -62,25 +62,27 @@
...
-#'   summary(pca)  # Print information about the factors
-#'   predict(pca, names=c("Component1", "Component2"))  # Back-predict scores
+#'   pca # Print loadings
+#'   summary(pca) # Print information about the factors
+#'   predict(pca, names = c("Component1", "Component2")) # Back-predict scores
 #'
 #'   # which variables from the original data belong to which extracted component?
 #'   closest_component(pca)
-#'
-#' \donttest{
```
strengejacke, line 80: See comment above

R/print.parameters_model.R
```diff
@@ -43,10 +43,21 @@
 #'   print(mp, select = c("Parameter", "Coefficient", "SE"))
 #'
 #'   print(mp, select = "minimal")
-#' }}
+#' }
```
strengejacke, line 46: Putting these curly braces into multiple lines adds empty rows into the examples from the PDF, thus the original formatting here is intentional.

R/print.parameters_model.R
```diff
@@ -64,7 +75,19 @@ print.parameters_model <- function(x, pretty_names = TRUE, split_components = TR
     insight::print_color("# Fixed Effects\n\n", "blue")
   }
 
-  formatted_table <- format(x, pretty_names = pretty_names, split_components = split_components, select = select, digits = digits, ci_digits = ci_digits, p_digits = p_digits, ci_width = "auto", ci_brackets = TRUE, format = "text")
+  formatted_table <-
```
strengejacke, line 78: see above (and other below, I don't comment on each repeating case)

R/robust_estimation.R
```diff
@@ -40,7 +40,8 @@
 #'     vcov_type = "CR2",
 #'     vcov_args = list(cluster = iris$cluster)
 #'   )
-#' }}
+#' }
+#' }
```
strengejacke, line 44: See above (and below)

strengejacke (changes requested):

R/factor_analysis.R
IndrajeetPatil: Oh, this is not something `styler` did. I did that because `styler` currently can't handle `donttest` nested inside `if` loop. So I just added everything to `donttest`. 

This is not so crucial, right? Even if the example is not tested, the tests are cover this function anyway.

IndrajeetPatil:
I will resolve the line break issue tomorrow. This is not done by `styler`. A lot of these changes are manual, like I went file by file to make these changes. So not everything is following a specific style guide. Though I do try to then run `styler` to follow the `tidyverse` style guide as much as possible.

R/model_parameters.psych.R
IndrajeetPatil: The examples are not run when wrapped in `donttest{}`, so this shouldn't change the behavior.

R/factor_analysis.R
strengejacke: I'm not sure `summary()` and `predict()` are included in tests, so we should either add a test or don't wrap those example parts in `\donttest{}`

R/select_parameters.R
```diff
@@ -54,7 +54,8 @@
 #'     iter = 500, refresh = 0, verbose = FALSE
 #'   )
 #'   select_parameters(model, cross_validation = FALSE)
-#' }}
+#' }
```
strengejacke, line 57: This should be `}}`?

strengejacke (approved):

strengejacke:
Looks like some examples were not formatted correctly: https://github.com/easystats/parameters/pull/352/checks?check_run_id=1467090676

PR #357, opened 2020-12-07
make `lavaan` and `lqmm` methods more robust

IndrajeetPatil:
If I write a wrapper function around `parameters::model_parameters`, it fails only for these two methods and so this PR makes them more robust to these edge cases:

``` r
# setup
set.seed(123)
library(lqmm)

# data
n <- 500
p <- 1:3 / 4
test <- data.frame(x = runif(n, 0, 1))
test$y <- 30 + test$x + rnorm(n)

# model
fit.lqm <-
  lqm(
    y ~ x,
    data = test,
    tau = p,
    control = list(verbose = FALSE, loop_tol_ll = 1e-9),
    fit = TRUE
  )

# plot
parameters::model_parameters(fit.lqm, eta_squared = NULL)
#> Error in boot.lqm(object, ...): unused argument (eta_squared = NULL)
```

<sup>Created on 2020-12-07 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>


strengejacke:
I would go a different route. Adding `fake`-arguments might raise confusion, so you instead might evaluate the dots, add valid arguments to a list (e.g. names `args`) and the pass it via `do.call()` in the function [`.extract_parameters_lavaan()`](https://github.com/easystats/parameters/blob/a187169724a57d1606dfa18dc02391b7c9da8c38/R/extract_parameters.R#L587).

```r
# here, evaluate dots instead of my example... make sure to keep only args 
# that are accepted by ?lavaan::parameterEstimates
args <- list(rsquare = TRUE, add.attributes = TRUE)
do.call(lavaan::parameterEstimates, c(list(object = fit, se = TRUE, level = .9), args))
```

strengejacke:
For `summary.lqmm()`, the `...` actually don't do much, so we might think about replacing

```r
cs <- summary(model, ...)
```

by

```r
cs <- summary(model)
```

in `.extract_parameters_lqmm()`.

IndrajeetPatil:
I think I'll leave `lavaan` out for now, seems a bit more complicated than `lqmm`.

strengejacke:
It's not that complicated, actually. All possible arguments are listed here:
https://rdrr.io/cran/lavaan/man/parameterEstimates.html

We have three arguments that we definitely want: `object = fit`, `se = TRUE` and `level = ci`. You can evaluate dots simply by:

```r
dot_args <- list(...)
```

This returns a named list, where names are argument names provides via `...`. You can than filter/match like:

```r
# list all argument names from the function, "..." is just a placeholder here
dot_args <- dot_args[names(dot_args) %in% c("zstat", "pvalue", "standardized", ...)] 
```

then do:

```r
data <- do.call(lavaan::parameterEstimates, c(list(object = fit, se = TRUE, level = ci), dot_args))
```

So it's literally adding just two lines of code (and replacing the last line shown here).

IndrajeetPatil:
The failing tests don't seem to have to do with anything I did.

strengejacke:
> The failing tests don't seem to have to do with anything I did.

No, it's an insight thing...

strengejacke:
Looks good! Need to submit parameters due to failing checks after effectsize update, so I merge this.

PR #364, opened 2020-12-12
WIP: split methods to package-specific files

vincentarelbundock:
This PR is marked as WIP because there `check()` produces one warning related to document (all tests are pass). I will try to figure it out soon.

Aside from that the project is essentially completed on my end, and the only changes I still intend to make (for now) are responses to PR review.

One thing to note is that by splitting methods over package-specific files, I had to use `@include` calls to make sure the documentation was collated in the right order. Once you know about this, it is rather easy.


vincentarelbundock:
Yes, I realize it is crazy to make a PR with 153 file changes. Sorry! (but it's kind of the nature of the proposed change...)

strengejacke:
I did not quite understand why we need the "Collate" field?

R/methods_AER.R
```diff
@@ -0,0 +1,63 @@
+# classes: .tobit, .gamlss
+
+# The `AER::ivreg` is being spun off to a separate package. The methods in 
+# `methods_ivreg.R` should work for objects produce by `AER`.
+
+
+#################### .tobit
+
+#' @include methods_gamlss.R
```
strengejacke, line 9: Why do we need this include tag here?

vincentarelbundock:
> I did not quite understand why we need the "Collate" field?

The PR still breaks, so I still need to figure out exactly how this works, but the basic problem is explained in this `roxygen2` doc:

https://rdrr.io/cran/roxygen2/man/update_collate.html

When building the docs, R will read file alphabetically. So it will start with `methods_AER.R`. But in that file, we assign `<-ci.lm`, which is later in `methods_lm.R`. When calling `devtools::document()`, this will produce an "object not found" error. Using `@include methods_lm.R` tells R to load the dependency first.

Again, it looks like it breaks right now, so I may not have gotten everything right just yet.

strengejacke:
Thanks, understand! The breaking checks came from another issue (should be fixed with https://github.com/easystats/parameters/pull/364/commits/50e1c04dcc07ff5c8bca23c7c356032587ac1c2c). Let's see if we need collate or not (I don't think so, your include-tags should be enough).

vincentarelbundock:
Sure, feel free to experiment. Note that the Collate field was created automatically by `roxygen2`. So if it works, you might have to delete every time you rebuild the docs.

strengejacke:
Ok, we indeed need the Collate field.

strengejacke:
Ok, checks look good! I'll now try to check if there are still duplicated functions, and maybe some minor additions (i.e. comments, like you did in the _AER file).

Thanks a lot for drafting this!

vincentarelbundock:
Fantastic. Let me know if you need help with any of that.

strengejacke:
My suggestion is that w try to stop to commit to _parameters_ master for now, and then we can check all the files over the next days, and then merge this PR. There is no hurry to do this, but don't worry if this PR is not merged the next two days.

vincentarelbundock:
seeing all the new commits, I have to say I'm feeling pretty bad for throwing you in the deep end like this. Hopefully the restructure isn't too bothersome and will bring some benefits in the future...

strengejacke:
Don't worry, the amount of commits make it look much more work than it actually is. Mostly, it's some minor moving of code, adding some comments, to harmonize the order of methods in the files, and this is done when I check the files for duplicated code.

I think I'm almost done, but will still look every now and then when I find some time. I think this is a great start for reorganizing the files. Later we can also move the remaining methods (like degrees_of_freedom()) if we like... But that's not that important, the mail goal was to have single files per package, so adding own methods is easier.

vincentarelbundock:
> Don't worry, the amount of commits make it look much more work than it actually is. Mostly, it's some minor moving of code, adding some comments, to harmonize the order of methods in the files, and this is done when I check the files for duplicated code.
> 
> I think I'm almost done, but will still look every now and then when I find some time. I think this is a great start for reorganizing the files. Later we can also move the remaining methods (like degrees_of_freedom()) if we like... But that's not that important, the mail goal was to have single files per package, so adding own methods is easier.

Cool. I might eventually do another round of changes for DOF et al. when I find time/energy.

Issue #321, opened 2020-10-21
draft for `model_parameters.PMCMR` 


IndrajeetPatil:
Will make a PR soon, unless you want something changed about the code.

``` r
# setup
set.seed(123)
library(PMCMRplus)

# custom function to convert from a matrix to a tidy dataframe
matrix_to_tidy <- function(m, col_name = "value") {
  result <-
    data.frame(
      group1 = rep(rownames(m), each = ncol(m)),
      group2 = rep(colnames(m), times = nrow(m)),
      col3 = as.numeric(base::t(m)),
      stringsAsFactors = FALSE
    )

  names(result)[3] <- col_name
  stats::na.omit(result)
}

# custom function
PMCMR_to_tidy <- function(mod, ...) {
  # combine all components of the object in a single dataframe
  cbind(
    merge(
      x = matrix_to_tidy(mod$statistic, col_name = "statistic"),
      y = matrix_to_tidy(mod$p.value, col_name = "p.value"),
      by = c("group1", "group2")
    ),
    data.frame(method = mod$method),
    data.frame(distribution = mod$dist),
    data.frame(p.adjust.method = mod$p.adjust.method)
  )
}

# example PMCMRplus object
ans1 <- kwAllPairsConoverTest(count ~ spray, data = InsectSprays)
#> Warning in kwAllPairsConoverTest.default(c(10, 7, 20, 14, 14, 12, 10, 23, : Ties
#> are present. Quantiles were corrected for ties.
PMCMR_to_tidy(ans1)
#>    group1 group2   statistic      p.value                   method distribution
#> 1       B      A   0.8897188 9.884327e-01 Conover's all-pairs test            q
#> 2       C      A -13.5821137 0.000000e+00 Conover's all-pairs test            q
#> 3       C      B -14.4718325 0.000000e+00 Conover's all-pairs test            q
#> 4       D      A  -8.8693844 4.607223e-07 Conover's all-pairs test            q
#> 5       D      B  -9.7591032 3.608281e-08 Conover's all-pairs test            q
#> 6       D      C   4.7127293 1.701980e-02 Conover's all-pairs test            q
#> 7       E      A -10.9546629 1.115604e-09 Conover's all-pairs test            q
#> 8       E      B -11.8443817 7.568990e-11 Conover's all-pairs test            q
#> 9       E      C   2.6274509 4.368688e-01 Conover's all-pairs test            q
#> 10      E      D  -2.0852785 6.814234e-01 Conover's all-pairs test            q
[...]

PR #381, opened 2020-12-31
adds support for PMCMR objects; #321

IndrajeetPatil:


strengejacke:
Thanks, should work now!

PR #389, opened 2021-01-03
closes #388

### LINKED ISSUE #388: renaming different CI columns for t-test outputs
It would be better here to rename `CI_low` and `CI_high` to `Difference_CI_low` and  `Difference_ CI_high`, resp., to make it more explicit what do these CIs correspond to, especially since the corresponding column is separated from them.

``` r
as.data.frame(parameters::model_parameters(
  t.test(1:10, y = c(7:20)),
  standardized_d = TRUE
))
#>   Parameter1 Parameter2 Mean_Parameter1 Mean_Parameter2 Difference        t
#> 1       1:10    c(7:20)             5.5            13.5         -8 -5.43493
#>   df_error   CI_low   CI_high Cohens_d  d_CI_low d_CI_high            p
#> 1 21.98221 -11.0528 -4.947198 -2.19089 -3.281465 -1.188038 1.855282e-05
#>                    Method
#> 1 Welch Two Sample t-test
```

<sup>Created on 2021-01-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

IndrajeetPatil:
Makes the column names more explicit


IndrajeetPatil:
Sorry, merging this without approval. Happy to later revert any changes that you may not like.

strengejacke:
The CI usually refers to the parameter estimate / point estimate, so I see no confusion for the "main" CI here. Thus, I would prefer to keep the column naming consistent, i.e. the "main" CI names would be `CI_low` and `CI_high`. This also makes the returned data frame consistent for all `model_parameter()` methods and the data (here: for the CI) easily accessible.

@mattansb @DominiqueMakowski Any thoughts?

DominiqueMakowski:
> The CI usually refers to the parameter estimate / point estimate, so I see no confusion for the "main" CI here

I agree

mattansb:
Also agree with Daniel

IndrajeetPatil:
Okay, I will revert the changes.

PR #392, opened 2021-01-04
add support for onesampb objects

IndrajeetPatil:


man/ci.merMod.Rd
```diff
@@ -1,16 +1,16 @@
 % Generated by roxygen2: do not edit by hand
 % Please edit documentation in R/2_ci.R, R/methods_DirichletReg.R,
-%   R/methods_betareg.R, R/methods_glmmTMB.R, R/methods_lme4.R,
-%   R/methods_mass.R, R/methods_mfx.R, R/methods_mixmod.R, R/methods_mixor.R,
+%   R/methods_MASS.R, R/methods_betareg.R, R/methods_glmmTMB.R,
```
strengejacke, line 3: As I mentioned somewhere else, please do a "fresh" pull from GitHub, so you have the file names in correct case (the new name is `methods_mass.R`, not `methods_MASS.R`), so this is not changed unintentionally...

strengejacke (approved):

man/ci.merMod.Rd
IndrajeetPatil: Oops, but what does it mean to do _fresh_ pull?

I always do `git pull upstream main` and then `git push`. 

How can I do a fresh pull?

strengejacke: I mean removing your parameters repo from hard disk and then just downbload/pull the repo again. Else, lower/upper case in filenames will be ignored (which is not crucial in this case, but might be in other cases).

IndrajeetPatil: Done!

PR #395, opened 2021-01-08
add effect size measure description for WRS2 objects

IndrajeetPatil:


IndrajeetPatil:
Failing tests seem to be due to the updated version of `MCMCglmm`: 
https://cran.r-project.org/web/packages/MCMCglmm/index.html

R/methods_wrs2.R
```diff
@@ -72,6 +73,7 @@ model_parameters.yuen <- function(model, verbose = TRUE, ...) {
       "df_error" = model$df,
       "p" = model$p.value,
       "Effsize" = model$effsize,
+      "Effectsize" = "Explanatory measure of effect size",
```
strengejacke, line 76: So what's the difference between "Effsize" and "Effectsize"? I would expect both to be the same, and I think we don't should have a column name "Effsize". It would be better to have two distinct names, and two proper names following the naming convention...

IndrajeetPatil: Yeah, I agree, but I am just staying faithful to the names `WRS2` itself uses:

```r
library(WRS2)
mod <- yuen(Anxiety ~ Group, data = spider)
str(mod)
#> List of 7
#>  $ test    : num 1.3
#>  $ conf.int: num [1:2] -17.93 4.43
#>  $ p.value : num 0.216
#>  $ df      : num 13.9
#>  $ diff    : num -6.75
#>  $ effsize : num 0.378
#>  $ call    : language yuen(formula = Anxiety ~ Group, data = spider)
#>  - attr(*, "class")= chr "yuen"
```

But I am happy to change it to `estimate` if you prefer.

strengejacke: Ok, then it would probably be "Coefficient"? I'm not sure what kind of model that is.

IndrajeetPatil: I don't think `"Coefficient"` is appropriate here, since there are estimates of effect sizes. I will change them to `Estimate`.

PR #404, opened 2021-01-23
begin to convert CIs to fractions

IndrajeetPatil:
I haven't been able to cover all the methods (e.g., `lm`, `glm`, `meta_bma`, etc., are still not covered here). 

Once you merge this, maybe you can see which methods are remaining where the CI needs to be converted to fraction.


R/methods_BayesFactor.R
```diff
@@ -153,6 +153,11 @@ model_parameters.BFBayesFactor <- function(model,
   attr(out, "pretty_names") <- pretty_names
   class(out) <- c("parameters_model", "see_parameters_model", class(out))
 
+  # converting CI to fraction
+  if ("CI" %in% names(out)) {
+    out$CI <- out$CI / 100
+  }
+
```
strengejacke, line 160: Here you have to coordinate with @DominiqueMakowski and _bayestestR_, else we just have 1% of the CI. Thus, you need to remove this code again and do all the CI-fraction-stuff in `bayestestR::describe_posterior()`.

R/methods_gam.R
```diff
@@ -67,6 +67,11 @@ model_parameters.gam <- function(model,
   attr(parameters, "object_name") <- deparse(substitute(model), width.cutoff = 500)
   class(parameters) <- c("parameters_model", "see_parameters_model", class(parameters))
 
+  # converting CI to fraction
+  if ("CI" %in% names(parameters)) {
+    parameters$CI <- parameters$CI / 100
+  }
+
```
strengejacke, line 74: This should actually be done in the related `ci()` method, which - if you caught all instances - should already return the fraction. To make maintenance easier and consistence, I would not add this `/ 100` code, but rather check that the `ci`-argument is properly set in the returned data frame.

R/methods_lme4.R
```diff
@@ -153,6 +152,11 @@ model_parameters.merMod <- function(model,
   attr(params, "object_name") <- deparse(substitute(model), width.cutoff = 500)
   class(params) <- c("parameters_model", "see_parameters_model", class(params))
 
+  # converting CI to fraction
+  if ("CI" %in% names(params)) {
+    params$CI <- params$CI / 100
+  }
+
```
strengejacke, line 159: See comment above.

R/methods_metaplus.R
```diff
@@ -94,6 +93,11 @@ model_parameters.metaplus <- function(model,
 
   attr(out, "title") <- unique(out$Method)
 
+  # converting CI to fraction
+  if ("CI" %in% names(out)) {
+    out$CI <- out$CI / 100
+  }
+
   out
```
strengejacke, line 101: See comment above.

strengejacke (changes requested):

strengejacke:
There are still quite a lot of tests that needs to be revised.

PR #406, opened 2021-01-26
bump package min versions; more 80-width styling

IndrajeetPatil:


IndrajeetPatil:
@strengejacke Does this look okay? I think you have already fixed the failing test in the `main` branch.

strengejacke:
When are the two workflows executed?

IndrajeetPatil:
- Testing code coverage and changes in it in PRs
- Checking for lints using `lintr`

IndrajeetPatil:
`lintr` workflow is currently failing because it can't find `M3C` package on `CRAN`.

strengejacke:
> `lintr` workflow is currently failing because it can't find `M3C` package on `CRAN`.

Yeah, that package is on Bioconductor.

strengejacke:
Hm, the workflows should either be fixed or removed, to avoid false positives.

IndrajeetPatil:
I've removed it.

strengejacke:
Same for test-coverage, though it would be great to have this workflow running.

strengejacke:
Maybe you can find a way how to prevent the workflow from quitting when there's a warning? Currently, warnings are converted to errors... Or how to install M3C at the right place.

IndrajeetPatil:
@strengejacke I have reverted all actions related changes, so no idea what is going on now.

PR #426, opened 2021-02-22
also re-export supported models function from insight

IndrajeetPatil:
Since `parameters` is more user-facing over `insight`, it will be a good to have a function that users can use to see if the model they want to use is supported or not.


strengejacke:
Actually, I wanted to get rid of the re-exports, they mostly exist to not break packages in the course of moving functions ;-)

IndrajeetPatil:
But is there any downside to this?

strengejacke:
No, not really. But maybe I'll remove some of the no longer necessary re-exports anyway, in the case they really don't fit

DominiqueMakowski:
Conceptually I don't really like re-exports either, it makes it confusing where a function is located (for people to debug or contribute) - it always bugged me in say rstan / rstanarm / brms that tend to re-export generously.

That said some day I wonder whether we should have all functions re-exported by "easystats" instead of loading the packages...

IndrajeetPatil:
> Conceptually I don't really like re-exports either, it makes it confusing where a function is located (for people to debug or contribute) - it always bugged me in say rstan / rstanarm / brms that tend to re-export generously.

Sure, but this is context-dependent. You make a good point for the functions which are user-facing, but the two functions in question are more developing-facing. I doubt there is someone out there trying to get to the source code of `standardize_names`. Most users will only be focusing on `model_parameters` function.

Another reason is that when I make  PRs to others' repos, I don't want to remove one dependency (`broom`) and add two more (`insight` and `parameters`) to their `Imports`. Number of package in `Imports` is something people now pay attention to since `CRAN` implemented a new check for excessive imports: https://github.com/wch/r-source/commit/0639b2c2f152834ac30bb5ed321e912706e34239

DominiqueMakowski:
> Another reason is that when I make PRs to others' repos, I don't want to remove one dependency (broom) and add two more (insight and parameters) to their Imports

Yes that's a good point, though it's only a virtual burden as we only depend on ourselves but it's true that it "looks" like a bad deal ^^ But then people need to also learn that a tree can hide a forest, like it's best to have 4 easystats dependencies that 1 other that itself has 20 of them... we can try to explain that

IndrajeetPatil:
Sure, but `_R_CHECK_EXCESSIVE_IMPORTS_` doesn't care about _recursive dependencies_, but only the count of packages in `Imports`.

IndrajeetPatil:
What's the final verdict on this? Can I merge or should I close this `PR`?

strengejacke:
I'd say merge.

Issue #436, opened 2021-03-08
allow parallel computation during bootstrapping


IndrajeetPatil:
This requires adding a new `parallel` argument to `model_parameters` and then passing the value to `boot` calls:

For example, here we can add `parallel = parallel` inside the call:

https://github.com/easystats/parameters/blob/d7fed242cc655d88c0417f17f48d15c657b67b2b/R/bootstrap_model.R#L85

We can also default to `parallel = "multicore"`, so multiple cores - if available -  are used by default.

DominiqueMakowski:
wouldn't it be better to let that be passed throug ellipsis to avoid cluttering the API? Or to retrieve it from the options (as stan does) ?

IndrajeetPatil:
Can't create a reprex because parallel doesn't seem to work with it. But passing the dots works (PR: #439).

```r
> set.seed(123)
> library(parameters)
> 
> mod <- lm(formula = wt ~ mpg, data = mtcars)
> 
> set.seed(123)
> system.time(model_parameters(mod, bootstrap = TRUE, iterations = 1000, parallel = "no")) 
   user  system elapsed 
  1.043   0.007   1.057 
> 
> set.seed(123)
> system.time(
+   model_parameters(
+     mod,
+     bootstrap = TRUE,
+     iterations = 1000,
+     parallel = "multicore",
+     ncpus = 4L
+   )
+ ) 
   user  system elapsed 
  0.078   0.056   0.613 
```

strengejacke:
"multicore" doesn't work on windows.

strengejacke:
Using normal R, or Microsoft R Open doesn't seem to make a difference, increasing used CPUs even slows down:

``` r
library(parameters)
#> Warning: Paket 'parameters' wurde unter R Version 4.0.4 erstellt
model <- lm(mpg ~ wt + cyl, data = mtcars)

microbenchmark::microbenchmark(
  model_parameters(model, bootstrap = TRUE, iterations = 1000, parallel = "snow", ncpus = 4),
  times = 5
)
#> Unit: seconds
#>                                                                                             expr
#>  model_parameters(model, bootstrap = TRUE, iterations = 1000,      parallel = "snow", ncpus = 4)
#>       min       lq    mean   median       uq      max neval
#>  2.146296 2.178574 2.18241 2.179772 2.200774 2.206634     5

microbenchmark::microbenchmark(
[...]

PR #439, opened 2021-03-09
stylistic changes

IndrajeetPatil:


IndrajeetPatil:
I will merge for now, and we can build on it further later.

strengejacke:
But checks don't pass.

IndrajeetPatil:
Uh-oh, let me check what happened.

strengejacke:
Looks like an invalid argument was used in the tests, which didn't do any harm, because we didn't pass down the dots to `boot()` ;-)
When we use boostrapping, dots may only contain valid arguments for `boot()`.

strengejacke:
@DominiqueMakowski This PR breaks a lot of other stuff, in particular all parameters that are passed down to `bayestestR::describe_posterior()`. Not sure what's the best way to go... Maybe we can capture the dots, then "filter" for valid arguments for `boot()`, and only pass those valid arguments to `boot()` (via `do.call()`).

@IndrajeetPatil would you revise this PR accordingly to not break current functionality?

DominiqueMakowski:
> Maybe we can capture the dots, then "filter" for valid arguments

wouldn't something like `parallel <- ifelse("parallel" %in% names(as.list(...)), as.list(...)$parallel, "no")` and then pass parallel to boot work?

IndrajeetPatil:
I prefer to then just go back to the earlier implementation I had in mind: 
explicitly providing `parallel` and `ncpus` argument, which, yes, complicates the API slightly, but would not cause any problems for any of the methods.

IndrajeetPatil:
As discussed in #436, I've decided to revert all `boot`-related changes. Now the PR only contains stylistic 80-width changes.

PR #444, opened 2021-03-17
Add the option to calculate Q1 and Q3

jmgirard:
I really like describe_parameters() but have missed the ability to calculate quartiles (25th and 75th percentiles), so I thought I'd try to add this as an optional (default = FALSE) argument.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -36,7 +36,12 @@ Authors@R:
            family = "Arel-Bundock",
            role = c("ctb"),
            email = "vincent.arel-bundock@umontreal.ca",
-           comment = c(ORCID = "0000-0003-1995-6531", Twitter = "@vincentab"))
+           comment = c(ORCID = "0000-0003-1995-6531", Twitter = "@vincentab")),
+    person(given = "Jeffrey",
+           family = "Girard",
+           role = c("ctb"),
+           email = "me@jmgirard.com",
+           comment = c(ORCID = "0000-0002-7359-3746", Twitter = "@jeffreymgirard"))
   )
 Description: Utilities for processing the parameters of various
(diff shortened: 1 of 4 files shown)


jmgirard:
Looks like I needed to do stats::quantile(), sorry.

strengejacke:
Thanks for this PR. Looks like some issues still needs to be resolved (documentation of argument?).

Don't forget to add yourself as contributor to the description file, if you like.

jmgirard:
Ok, I'll work on it. Still new to collaborating on shared repos so please be patient with me.

IndrajeetPatil:
I think you just need to run `roxygen2::roxygenize()`, so that the docs (`.Rd` files) are also updated to contain the new argument.

strengejacke:
I just realized that there was a bug (in the current parameters, not due to your PR, but I realized that only by reviewing this PR).

strengejacke:
Great, thanks! Is there anything you want to add/modify? Else, this is ready for merging.

jmgirard:
I think we're good, thanks!

IndrajeetPatil:
@strengejacke We need to update print method to also return `Quartiles` column in brackets, right?

Screenshot 2021-03-18 at 15 23 08

strengejacke:
@jmgirard used the same formatting (i.e. brackets), but I changed that and removed the brackets. I think _brackets_ (or parenthesis) indicate an _interval_ (or a range), while the quartiles here indicate two cut-off points. The quartiles do not range from (e.g.) 5.10 to 6.40. Thus, I thought removing brackets would make more sense.

IndrajeetPatil:
Oh, good point! 

Yeah, that makes sense.

PR #456, opened 2021-03-30
checking if snapshot test fails on CI

IndrajeetPatil:


strengejacke:
Can you make sure that all "invalid" (sub)directories created by snapshot tests are excluded from building the source file? And skip that test on CRAN.

IndrajeetPatil:
> Can you make sure that all "invalid" (sub)directories created by snapshot tests are excluded from building the source file?

So just add `_snaps` folders to `.Rbuildignore`?

strengejacke:
> > Can you make sure that all "invalid" (sub)directories created by snapshot tests are excluded from building the source file?
> 
> So just add `_snaps` folders to `.Rbuildignore`?

Yes, I would suggest doing that, just to be safe.

PR #461, opened 2021-04-02
test only lqm and not lqmm

IndrajeetPatil:


strengejacke:
I think it mainly was because you used `df_lqmm$Coefficient` instead of `df_lqmm`

PR #469, opened 2021-04-05
link to effectsize vignette

IndrajeetPatil:
Following discussion here: https://github.com/easystats/easystats/issues/112#issuecomment-813264888


strengejacke:
I don't see much "overlap" in terms of redundancy. The effectsize vignette is much more technical, describing the parallels to other effect size coefficients, while the parameters version of standardizing model coefficients is rather showing the different options, the differences in terms of interpretation and thus is much closer to an "application perspective".

PR #499, opened 2021-05-20
Resolve #491

### LINKED ISSUE #491: check_sphericity in parameters and performance
@mattansb since you implemented this here (and @DominiqueMakowski in _parameters_), could you check whether we can merge these functions (and decide on a package where to put this function into):

https://easystats.github.io/performance/reference/check_sphericity.html
https://easystats.github.io/parameters/reference/check_sphericity.html

mattansb:
> The problem is that {parameters} and {performance} don't import each other, so where would the check_sphericity <- function(x, ...) UseMethod("check_sphericity") function live without getting masking warnings when loading both packages? AFAIK we also can't reexport a function in a situation like this, no?


strengejacke:
I think we need to find another solution than re-exporting, because currently, _parameters_ does not depend on _performance_.

mattansb:
@strengejacke I've just made it *yes* import now...

So we either have -
- **Reexport** + add performance as dep
- **Rename** the function
- **Remove** the function (and `check_kmo()`) and use them only as methods to `check_factorstructure()`.

Cast your vote? @DominiqueMakowski @bwiernik

mattansb:
So we either have -
- **Reexport** + add performance as dep
    - or not just add the function, and have a masking warning... (do we care?)
- **Rename** the function
- **Remove** the function (and `check_kmo()`) and use them only as methods to `check_factorstructure()`.

Cast your vote? @DominiqueMakowski @bwiernik

mattansb:
@strengejacke Ready to merge!

PR #539, opened 2021-06-23
remove unnecessary re-exports; get rid of `NOTE`s

IndrajeetPatil:
@strengejacke  There is one `NOTE` I couldn't get rid of:

```r
print_html.compare_parameters: no visible binding for global variable
    ‘table_caption’
  Undefined global functions or variables:
    table_caption
```

You can see into it later.


bwiernik:
@IndrajeetPatil  line 137 of print_html.R. There is no assignment of `table_caption` inside `print_html.compare_parameters()`

strengejacke:
I'm not sure you can safely remove all these re-exports at the moment.

IndrajeetPatil:
I think removing them will be the easiest way to find all instances where they are being used. 

Unfortunately, we will have to rip this band aid off in one go at some point. It might as well be now.

strengejacke:
I have already checked some functions, and I think we need all those that were available before this PR.

IndrajeetPatil:
But shouldn't the solution be then to use `datawizard` imports in those functions, instead of continuing to re-export these functions from `parameters`?

I have already done this in `correlation`, `modelbased`, and `parameters`. This is not relevant for `insight` and `datawizard`.
Mattan has done this for `effectsize`.

So what remains is checking the same thing for the remaining packages, like`see`, `performance`, etc.
I can do that.

strengejacke:
yeah, but it requires those packages being on CRAN first, and I'm not sure in which order we submit to CRAN. Furthermore, there are also packages outside the easyverse that rely on re-exported functions, which then all would require to be updated to CRAN before the next parameters update.

IndrajeetPatil:
Okay, all the `datawizard` functions are once again re-exported.

PR #597, opened 2021-09-19
requiet: testthat log readability

vincentarelbundock:
Same PR as my recent one on `insight`. Goal is to make the `testthat` log easier to read by suppressing package startup messages.


vincentarelbundock:
I believe that the same errors are raised on `master`, and that they are unrelated to this PR.

strengejacke:
Thanks a lot!

PR #604, opened 2021-10-03
tinyverse badge

vincentarelbundock:
For me, the main benefit of `parameters` over competitors is that it has virtually no dependencies. This PR adds [a tinyverse badge](https://www.tinyverse.org/) to showcase that great benefit on the README.

I only added one line; the other changes to `README.md` were generated automatically by `rmarkdown::render`


IndrajeetPatil:
Sounds like a good idea!

Out of curiosity, how is one to read this badge? 
`{parameters}` has 2 hard dependencies (`{bayestestR}` and `{insight}`), but now sure what `3` means in the badge.

image

vincentarelbundock:
The number after the slash is recursive. It includes all imports, and also all the imports of your imports, etc.

So the key number is the second one.

Issue #639, opened 2021-12-21
Confidence intervals mis-aligned for glmmTMB


d-morrison:
Here's a reprex:

```
library(parameters)
library(glmmTMB)

load(url("https://github.com/d-morrison/parameters/raw/glmmTMB/data/pressure_durations.RData"))

pressure1 = glmmTMB(
  family = nbinom2,
  data  = pressure_durations,
  formula = n_samples ~ Surface + Side + Jaw + (1 | Participant / Session),
  ziformula = ~ Surface + Side + Jaw + (1 | Participant / Session),
  dispformula = ~ 1)

parameters(pressure1, component = "zi", effects = "random")
#> # Random Effects (Zero-Inflated Model)
#> 
#> Parameter                           | Coefficient |       95% CI
#> ----------------------------------------------------------------
#> SD (Intercept: Session:Participant) |        0.69 | [0.40, 1.19]
#> SD (Intercept: Participant)         |        2.39 | [0.40, 1.19]
#> 
#> Uncertainty intervals (equal-tailed) and p values (two-tailed) computed using a
#>   Wald z-distribution approximation.
```
The confidence interval for the Participant-level SD appears to be accidentally copied from the Session:Participant level SD. The correct confidence intervals are given by confint:

```
confint(pressure1, parm = "theta_")[3:4,]
#>                                                2.5 %   97.5 %  Estimate
#> Session:Participant.zi.Std.Dev.(Intercept) 0.3968345 1.187371 0.6864327
#> Participant.zi.Std.Dev.(Intercept)         1.2544186 4.566486 2.3933837
```
I tracked the error into .random_sd_ci(), where the line pattern <- paste0("^(zi\\.|", group_factor, "\\.zi\\.)") appears to have trouble when group_factor has more than one element. It seems like that regexp (and another one a few lines down) would need to be modified to look for either grouping level. I have attempted a fix (will PR shortly).

strengejacke:
closed in #640

PR #640, opened 2021-12-21
glmmTMB confidence intervals realigned

d-morrison:
See issue #639; I have attempted to correct the issue.


IndrajeetPatil:
Thanks.

@strengejacke can review this, but I already had a couple of comments:

- In `{easystats}`, we try to keep the hard dependencies to 0. So you will need to refactor your code to remove dependence on `{dplyr}`. 
- The base pipe (`|>`) was introduced in `R 4.0`, but we support R versions `3.4` onwards, so you also can't use it in your code.

d-morrison:
> Thanks.
> 
> @strengejacke can review this, but I already had a couple of comments:
> 
> * In `{easystats}`, we try to keep the hard dependencies to 0. So you will need to refactor your code to remove dependence on `{dplyr}`.
> * The base pipe (`|>`) was introduced in `R 4.0`, but we support R versions `3.4` onwards, so you also can't use it in your code.

I think I switched all the edits to base R syntax; hope that helps!

strengejacke:
Thanks for the PR! I'll take a closer look today or the next days.

strengejacke:
Thanks, this looks good! Will wait for the checks, I think some tests might be adopted because they might contain the wrong output from `print()`.

PR #645, opened 2022-01-08
Close #644

### LINKED ISSUE #644: `model_parameters.merMod` is very slow
The `parameters` function is extremely slow when applied to a model produced by the `lme4::lmer` function. In this example, `broom.mixed` is basically instantaneous but `parameters` quite slow. Adding more random effects units (e.g., 10000) makes things much worse; `parameters` often seems to hang indefinitely and is unuseable.

``` r
library(lme4)
library(tictoc)
library(broom.mixed)
library(parameters)

N <- 1e6
dat <- data.frame(
    x = rnorm(N),
    y = rnorm(N),
    k = sample(1:50, N, replace = TRUE),
    m = sample(1:1000, N, replace = TRUE))
mod <- lmer(y ~ x + (1 | k) + (1 | m), data = dat)

tic()
tidy(mod)
```

    ## # A tibble: 5 × 6
    ##   effect   group    term             estimate std.error statistic
    ##   <chr>    <chr>    <chr>               <dbl>     <dbl>     <dbl>
    ## 1 fixed    <NA>     (Intercept)      0.000787  0.00128      0.616
    ## 2 fixed    <NA>     x               -0.000741  0.000998    -0.743
    ## 3 ran_pars m        sd__(Intercept)  0.00369  NA           NA    
    ## 4 ran_pars k        sd__(Intercept)  0.00557  NA           NA    
    ## 5 ran_pars Residual sd__Observation  0.998    NA           NA

``` r
toc()
```

    ## 0.179 sec elapsed

``` r
tic()
parameters(mod)
```

    ## # Fixed Effects
    ## 
    ## Parameter   | Coefficient |       SE |        95% CI | t(999995) |     p
    ## ------------------------------------------------------------------------
    ## (Intercept) |    7.87e-04 | 1.28e-03 | [ 0.00, 0.00] |      0.62 | 0.538
    ## x           |   -7.41e-04 | 9.98e-04 | [ 0.00, 0.00] |     -0.74 | 0.457
    ## 
    ## # Random Effects
    ## 
    ## Parameter         | Coefficient
    ## -------------------------------
    ## SD (Intercept: m) |    3.69e-03
    ## SD (Intercept: k) |    5.57e-03
    ## SD (Residual)     |        1.00

    ## 
    ## Uncertainty intervals (equal-tailed) and p values (two-tailed) computed using a
    ##   Wald t-distribution approximation.

``` r
toc()
```

    ## 10.908 sec elapsed

Here is a screenshot of some profiling:

profile

strengejacke:
Re-implement core parts of `insight::get_variance()`, in a simpler form to speed up computations of random variances.

### DIFF
--- a/R/extract_random_variances.R
+++ b/R/extract_random_variances.R
@@ -482,6 +482,254 @@
 
 
 
+
+
+
+# Extract Variance and Correlation Components ----
+
+# store essential information about variance components...
+# basically, this function should return lme4::VarCorr(x)
+.get_variance_information <- function(x, model_component = "conditional") {
+
+  # reason to be installed
+  reason <- "to compute random effect variances for mixed models"
+
+  # installed?
+  insight::check_if_installed("lme4", reason = reason)
+
+  if (inherits(x, "lme")) {
+    insight::check_if_installed("nlme", reason = reason)
+  }
+
+  if (inherits(x, "clmm")) {
+    insight::check_if_installed("ordinal", reason = reason)
+  }
+
+  if (inherits(x, "brmsfit")) {
+    insight::check_if_installed("brms", reason = reason)
+  }
+
+  if (inherits(x, "cpglmm")) {
+    insight::check_if_installed("cplm", reason = reason)
+  }
+
+  if (inherits(x, "rstanarm")) {
+    insight::check_if_installed("rstanarm", reason = reason)
+  }
+
+  # stanreg
+  # ---------------------------
+  if (inherits(x, "stanreg")) {
+    varcorr <- lme4::VarCorr(x)
+
+    # GLMMapdative
+    # ---------------------------
+  } else if (inherits(x, "MixMod")) {
+    vc1 <- vc2 <- NULL
+    re_names <- insight::find_random(x)
+
+    vc_cond <- !grepl("^zi_", colnames(x$D))
+    if (any(vc_cond)) {
+      vc1 <- x$D[vc_cond, vc_cond, drop = FALSE]
+      attr(vc1, "stddev") <- sqrt(diag(vc1))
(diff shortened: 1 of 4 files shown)


vincentarelbundock:
This sounds like a good idea, but I'm afraid I don't know these models well enough to offer serious code review.

That said, I did install the PR branch and ran some profiling. The good news is that there now seems to be very little overhead in the `.get_variance_info`: virtually all of its time is spent calling `ranef()`.

The bad news is that `.get_variance_info` is called 3 times(!), so the overall time it takes to execute `model_parameters`  has not come down much.

bwiernik:
Can we consolidate that to calling it just once?

vincentarelbundock:
> Can we consolidate that to calling it just once?

Or even make it entirely optional and non-default, as in `broom.mixed`: https://github.com/bbolker/broom.mixed/blob/master/R/lme4_tidiers.R#L287

(I'm not necessarily advocating this; just curious about the pros and cons.)

strengejacke:
I haven't implemented any changes that affects the code, I just copied the helper functions, so nothing should have changed according to performance. 😬

Will continue to work on it tomorrow.

The idea is that retrieving the varcorr is only called once instead of three or more times, and ranef is no longer needed.

strengejacke:
> > Can we consolidate that to calling it just once?
> 
> Or even make it entirely optional and non-default, as in `broom.mixed`: https://github.com/bbolker/broom.mixed/blob/master/R/lme4_tidiers.R#L287
> 
> (I'm not necessarily advocating this; just curious about the pros and cons.)

I remember s discussion that suggested including these by default. And I would say if we can significant reduce the time to calculate the variance components, it can be reported by default?

PR #646, opened 2022-01-11
update to v2 for pkgdown GHA

IndrajeetPatil:


strengejacke:
It's not the circular dependency, parameters is suggesting a non existing effectsize version

strengejacke:
I should reduce the version number to 0.5.0

IndrajeetPatil:
Ah, I see. `effectsize 0.5.1` is still not merged in the `main` branch.

Yeah, let's see if that changes things.

mattansb:
As long as the actual code and test require a higher version of effectsize, we should be good.

PR #656, opened 2022-01-28
use insight helpers

IndrajeetPatil:
https://github.com/easystats/easystats/issues/205

Wait for insight update on CRAN before merging

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -176,7 +176,7 @@ Suggests:
     VGAM,
     WRS2
 Remotes:
-    easystats/datawizard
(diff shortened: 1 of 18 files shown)


IndrajeetPatil:
@strengejacke I think the insight versions of helpers are doing something different from the ones in parameters and performance. So we will need to keep having separate copies across these packages.

strengejacke:
They shouldn't. Maybe we need a default method in insight that just returns x. And we should set verbose to FALSE here, or default it to FALSE in insight. Will look later

PR #663, opened 2022-02-22
Bugfix in .extract_random_variances()

bwiernik:
Currently, models with different grouping factors for intercepts vs slopes fail to extract variances

### DIFF
--- a/R/extract_random_variances.R
+++ b/R/extract_random_variances.R
@@ -157,16 +157,14 @@
     ran_intercept$Group <- rownames(ran_intercept)
     ran_intercept$Parameter <- "SD (Intercept)"
   }
-  ran_groups <- ran_intercept$Group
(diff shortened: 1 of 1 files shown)


R/extract_random_variances.R
```diff
@@ -175,8 +173,12 @@
         ran_slope$Group[slopes] <- i
       }
     }
+  } else {
+    rand_groups_slp <- NULL
```
strengejacke, line 177: typo here, should be `ran_groups_slp`.

strengejacke (changes requested):

R/extract_random_variances.R
bwiernik, line 177:
```suggestion
    ran_groups_slp <- NULL
```

PR #672, opened 2022-03-07
move `demean` vignette from `{datawizard}`

IndrajeetPatil:
https://github.com/easystats/datawizard/issues/91


strengejacke:
I think you should also update the pkgdown.yml.

PR #690, opened 2022-04-01
model_parameters supports posterior-pkg

strengejacke:


strengejacke:
@DominiqueMakowski @mattansb see https://github.com/easystats/parameters/pull/690/files#diff-48fbe95ae07f81675eb69a1f581e191e8cff693327702465a5bcff30390ad2f9R50-R73

We could use these helpers in _bayestestR_, too, and support the _posterior_ package in _bayestest_R. Looks like simply having a `.draws` class for each method, that converts input to data frame and then calls the related data-frame-method.

strengejacke:
Need to revise docs for `?model_parameters.draws`, the `model` argument still states that a data frame is allowed.

bwiernik:
I think we should probably have an as_draws arguments in the data.frame method that passes it to the draws method instead. For handling of manual bootstrapped data frames

strengejacke:
Defaults to `FALSE`, or?

bwiernik:
Yes, default to FALSE. That ensures there is a quick fix for folks who relied on the previous behavior

PR #691, opened 2022-04-02
use get_varcov directly for robust VCOV

strengejacke:
@vincentarelbundock Not sure, but I think it's really that easy to switch to `get_varcov()` for robust estimation, right?


vincentarelbundock:
Yep, that was my big master plan :)

Crossing fingers...

vincentarelbundock:
You also need to require the very latest version of `insight`

strengejacke:
Yes, I think the insight-GH should already be used via "Remotes" in the DESCRIPTION file.

strengejacke:
passed the wrong argument names, that's why tests were failing... What was the reason why we used `do.call()` here? I now call `insight::get:varcov()` directly.

vincentarelbundock:
Commit https://github.com/easystats/insight/commit/bbdb4808567806a942ea3e59d47f8f2dc490d044 from `insight` removes a lot of the warnings on previous testing runs.

Unfortunately, I probably won't have time to look more at this in the very short run.

bwiernik:
This looks good to me.

vincentarelbundock:
same. looks simple enough, and the tests now look like they are passing.

strengejacke:
merge now, or after _insight_ was updated on CRAN again?

PR #692, opened 2022-04-04
use eti instead hdi by default

strengejacke:
See discussion at https://github.com/easystats/bayestestR/issues/521 and http://www.stat.columbia.edu/~gelman/research/published/spin.pdf


bwiernik (approved):

strengejacke:
Can anybody explain why this test is failing on win-devel for some time?
https://github.com/easystats/parameters/runs/5812853125?check_suite_focus=true#step:10:403

@IndrajeetPatil @mattansb @bwiernik @DominiqueMakowski

IndrajeetPatil:
If I am reading the error correctly, the following extra line is printed only on windows R-devel?

> "r2: 0.743; ar2: 0.613; wr2: 0.180; war2: 0.175"

I don't see any relevant change in [NEWS](https://cran.r-project.org/bin/windows/base/NEWS.R-devel.html) that might explain this. 

Unless _fixest_ is itself behaves in unexpected ways on windows-devel. I don't see any of its tests failing in CRAN daily checks, but then we can't also be sure if this is something that is tested in the package. 

TLDR: No idea why 😟

strengejacke:
The strange thing is that this line should be omitted in general, by `out[-9]`. So it seems some of the other output must have changed, and I also can't see which one (and parameters does not fail with this test on CRAN - yet)

IndrajeetPatil:
A temporary solution would be to skip this test on `windows-devel` and add a `# FIXME` tag.

strengejacke:
I think I found the source of this error. The warning "the condition has length > 1 and only the first element will be used" will probably become an error in a future release, and for R devel (R 4.3, since a few days), this is now the case. That's why this error occurred just recently, because R devel is now 4.3, no longer 4.2 - and it seems the switch from warning to error comes in 4.3

``` r
library(datawizard)
library(parameters)
library(fixest)
#> 
#> Attaching package: 'fixest'
#> The following objects are masked from 'package:parameters':
#> 
#>     demean, dof
#> The following object is masked from 'package:datawizard':
#> 
#>     demean

data("qol_cancer")
qol_cancer <- cbind(
  qol_cancer,
  datawizard::demean(qol_cancer, select = c("phq4", "QoL"), group = "ID")
)

m <- feols(
  QoL ~ time + phq4 | ID,
  data = qol_cancer
)

fixest::r2(m)
#> Warning in if (is_calling_fun("deltaMethod")) {: the condition has length > 1
#> and only the first element will be used

#> Warning in if (is_calling_fun("deltaMethod")) {: the condition has length > 1
#> and only the first element will be used
#>       cor2         r2        ar2        pr2       apr2        wr2       war2 
#> 0.74313219 0.74313219 0.61332466 0.15417324 0.07815105 0.17985575 0.17546995 
#>       wpr2      wapr2 
#> 0.02590105 0.02497458
```

<sup>Created on 2022-04-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

strengejacke:
Same code run on R devel:

``` r
library(datawizard)
library(parameters)
library(fixest)
#> 
#> Attaching package: 'fixest'
#> The following objects are masked from 'package:parameters':
#> 
#>     demean, dof
#> The following object is masked from 'package:datawizard':
#> 
#>     demean

data("qol_cancer")
qol_cancer <- cbind(
  qol_cancer,
  datawizard::demean(qol_cancer, select = c("phq4", "QoL"), group = "ID")
)

m <- feols(
  QoL ~ time + phq4 | ID,
  data = qol_cancer
)

fixest::r2(m)
#> Error in if (is_calling_fun("deltaMethod")) {: the condition has length > 1
```

<sup>Created on 2022-04-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

IndrajeetPatil:
What do you use to switch between different versions of R? Manually, or do you have a professional version of RStudio that does allow this?

bwiernik:
You're on Mac I take it? On windows, you can have multiple R versions installed and swap between them in RStudio

strengejacke:
@IndrajeetPatil I have the standard (free) RStudio, and I switch by changing the R version in the global options. You just need to install multiple versions (and probably install all packages again)

IndrajeetPatil:
Thanks, both!

Just to be sure, none of you see this in Global Options for your free RStudio versions, right?

image

Source: https://support.rstudio.com/hc/en-us/articles/212364537-Using-multiple-versions-of-R-with-RStudio-Workbench-RStudio-Server-Pro#:~:text=You%20can%20select%20different%20versions,workspace%20before%20restarting%20your%20session.

bwiernik:
There is a similar option in the free version of Windows RStudio. Not on Mac

IndrajeetPatil:
Wow! There indeed is. I don't know how I missed this?! I have both Windows (work) and Mac (personal) machines, so can switch depending on if I need multiple versions. 

This is so cool! Thanks both.

PR #695, opened 2022-04-17
add `sort_parameters()` function

IndrajeetPatil:
Example:

``` r
library(parameters)

# creating object to sort (can also be a regular data frame)
mod <- model_parameters(stats::lm(wt ~ am * cyl, data = mtcars))

# original output
mod
#> Parameter   | Coefficient |   SE |        95% CI | t(28) |      p
#> -----------------------------------------------------------------
#> (Intercept) |        1.66 | 0.59 | [ 0.46, 2.86] |  2.82 | 0.009 
#> am          |       -0.96 | 0.79 | [-2.58, 0.67] | -1.21 | 0.238 
#> cyl         |        0.30 | 0.08 | [ 0.13, 0.47] |  3.68 | < .001
#> am * cyl    |        0.03 | 0.13 | [-0.23, 0.30] |  0.25 | 0.803
#> 
#> Uncertainty intervals (equal-tailed) and p-values (two-tailed) computed using a
#>   Wald t-distribution approximation.

# sorted outputs
sort_parameters(mod, sort = "ascending")
#> Parameter   | Coefficient |   SE |        95% CI | t(28) |      p
#> -----------------------------------------------------------------
#> am          |       -0.96 | 0.79 | [-2.58, 0.67] | -1.21 | 0.238 
#> am * cyl    |        0.03 | 0.13 | [-0.23, 0.30] |  0.25 | 0.803 
#> cyl         |        0.30 | 0.08 | [ 0.13, 0.47] |  3.68 | < .001
#> (Intercept) |        1.66 | 0.59 | [ 0.46, 2.86] |  2.82 | 0.009
#> 
#> Uncertainty intervals (equal-tailed) and p-values (two-tailed) computed using a
#>   Wald t-distribution approximation.

sort_parameters(mod, sort = "descending")
#> Parameter   | Coefficient |   SE |        95% CI | t(28) |      p
#> -----------------------------------------------------------------
#> (Intercept) |        1.66 | 0.59 | [ 0.46, 2.86] |  2.82 | 0.009 
#> cyl         |        0.30 | 0.08 | [ 0.13, 0.47] |  3.68 | < .001
#> am * cyl    |        0.03 | 0.13 | [-0.23, 0.30] |  0.25 | 0.803 
#> am          |       -0.96 | 0.79 | [-2.58, 0.67] | -1.21 | 0.238
#> 
#> Uncertainty intervals (equal-tailed) and p-values (two-tailed) computed using a
#>   Wald t-distribution approximation.
```

<sup>Created on 2022-04-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


strengejacke:
See check issues

strengejacke:
Nice idea! Maybe we can add a method for `parameters_model` class, and the default class checks if input is a model, and if so, first call `model_parameters()` and then `sort_parameters()`. What do you think?

IndrajeetPatil:
> Nice idea! Maybe we can add a method for parameters_model class, and the default class checks if input is a model, and if so, first call model_parameters() and then sort_parameters(). What do you think?

I don't know. I feel like it might be an overkill. The context in which I was seeing the utility of this function was where users/developers have already tidied the output (either using _easystats_ or _broom_) and wish to sort it only for visualization or for tables, etc. I am trying to adopt [SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) as much as possible for these utility functions, so it'd be a bit unexpected for a function meant for sort to also tidy inputs.

One thing we can do is to add a new `sort` parameter to `model_parameters` signature, so that users don't even need to call `sort_parameters()`. But I'd still like to have this function because it can handle the more general case where tidiying happened outside _easystats_.

Sounds reasonable to you?

IndrajeetPatil:
@strengejacke This should be all set.

See my comment above and also https://github.com/easystats/parameters/issues/696.

strengejacke:
Yes, sounds reasonable.

IndrajeetPatil:
Can I merge this?

PR #702, opened 2022-04-30
move standardize parameters from effectsize to parameters

strengejacke:


strengejacke:
Anybody an idea why this error occurs?
https://github.com/easystats/parameters/runs/6247401150?check_suite_focus=true#step:10:496

bwiernik:
That's bizarre. Does it need to be stats::gaussian()?

PR #711, opened 2022-05-11
fix `p_value.numcric()`

mattansb:
For an approx. bootstrap p-value.


mattansb:
@strengejacke Seems good to merge (:

PR #718, opened 2022-05-30
model_parameters docs

vincentarelbundock:
Improve docs for the `model_parameters()` function:

1. All supported modeling packages are explicitly listed under one of the categories
2. Consolidated some related categories
3. Explicitly tell users to "follow a link". This is especially for users -- like me -- who don't use HTML documentation and won't automatically "see" the links.

The new docs look like this:

Screenshot 2022-05-30 083632


strengejacke:
I think some of the check issues arise from the fact that rd-files need to be collated in the description (or namespace?). I started renaming files so they appear in the necessary alphabetical order, so `@rdname` etc. work as expected. The alphabetical order avoids tons of "collate" entries in the description file, but is of course a bit more complicated to take care of where which arguments are documented.

vincentarelbundock:
I'll investigate.

But FWIW, I think there's essentially zero cost in having collate entries in DESCRIPTION. They're at the bottom of the file and noone reads until there. In contrast, messing with file names to hope that the order is *just* right feels like a hassle.

PR #720, opened 2022-05-31
Simply extracting random effects variances

### LINKED ISSUE #717: Simply extracting random effects variances
We could probably simplify a lot for merMod and glmmTMB objects:

```r
m <- lme4::lmer(angle ~ poly(temp, 2) + (poly(temp, 2)|replicate) + (temperature|recipe),
                data = lme4::cake)
#> boundary (singular) fit: see help('isSingular')

as.data.frame(lme4::VarCorr(m), order = "lower.tri")
#>          grp           var1           var2          vcov       sdcor
#> 1  replicate    (Intercept)           <NA>  39.289063184  6.26809885
#> 2  replicate    (Intercept) poly(temp, 2)1  42.565947869  0.62190190
#> 3  replicate    (Intercept) poly(temp, 2)2  17.810281023  0.32657990
#> 4  replicate poly(temp, 2)1           <NA> 119.236487929 10.91954614
#> 5  replicate poly(temp, 2)1 poly(temp, 2)2  89.615008219  0.94325840
#> 6  replicate poly(temp, 2)2           <NA>  75.699143015  8.70052544
#> 7     recipe    (Intercept)           <NA>   0.629862937  0.79363905
#> 8     recipe    (Intercept)  temperature.L  -0.108512942 -1.00000000
#> 9     recipe    (Intercept)  temperature.Q  -0.218140798 -0.95405846
#> 10    recipe    (Intercept)  temperature.C  -0.973441126 -0.99998229
#> 11    recipe    (Intercept)  temperature^4  -1.384011822 -0.99823861
#> 12    recipe    (Intercept)  temperature^5  -0.018849548 -0.09504993
#> 13    recipe  temperature.L           <NA>   0.018694636  0.13672833
#> 14    recipe  temperature.L  temperature.Q   0.037581350  0.95405846
#> 15    recipe  temperature.L  temperature.C   0.167704676  0.99998229
#> 16    recipe  temperature.L  temperature^4   0.238437897  0.99823861
#> 17    recipe  temperature.L  temperature^5   0.003247405  0.09504993
#> 18    recipe  temperature.Q           <NA>   0.082999939  0.28809710
#> 19    recipe  temperature.Q  temperature.C   0.337762423  0.95582435
#> 20    recipe  temperature.Q  temperature^4   0.470380117  0.93460406
#> 21    recipe  temperature.Q  temperature^5  -0.014942360 -0.20756504
#> 22    recipe  temperature.C           <NA>   1.504487939  1.22657570
#> 23    recipe  temperature.C  temperature^4   2.138207522  0.99786799
#> 24    recipe  temperature.C  temperature^5   0.027316308  0.08912541
#> 25    recipe  temperature^4           <NA>   3.051861461  1.74695777
#> 26    recipe  temperature^4  temperature^5   0.067198680  0.15394027
#> 27    recipe  temperature^5           <NA>   0.062438464  0.24987690
#> 28  Residual           <NA>           <NA>  21.566909051  4.64401863
```

If `var2` is given, it's a correlation, if `var2` is `NA`, it's a SD.

_Originally posted by @strengejacke in https://github.com/easystats/parameters/issues/716#issuecomment-1140758184_

strengejacke:
Close #717


strengejacke:
ok, two things:

1. looks like there are some differences in rounding between the old and the new approach to extract random effects variances.
2. looks like the order of parameters changed in a few places. this means, either fixing this in the code, or - if the new sorting is OK - fix the related tests.

mattansb:
I don't think the order of terms should matter too much...

strengejacke:
I fixed the ordering-issue now, and actually improved it. Order might matter, because in your poly-examples, parameters were ordered alphabetically (.C, .L, .Q), which should be .L, .Q and .C (and then .4 and .5). This was previously a wrong behaviour and should now also be fixed.

This PR still needs some testing, especially for profiled-CIs, to see if parameters and SE/CI still match. Feel free to play around with some models using `model_parameters(model, ci_method = "profile")` and see if the CIs still match those from `ci_method = "wald"` (they're not identical, but I think we can see if the CIs are "matching" and in the correct order).

strengejacke:
And we need to check if this also work for other packages, like GLMMadaptive, ordinal:clmm, nlme etc.

mattansb:
> you see that correlations are not bounded between [-1, 1]. This is something to be fixed in bayestestR, I think, but do you have any idea how to transform the posterior in order to have a proper "point estimate" here?

I think the values from `m1` are covariances and variances, not correlations and standard deviation. So you would have to:

```R
library(rstanarm)
library(lme4)
data(cake)
set.seed(123)

m1 <- stan_lmer(angle ~ temp + (temp | replicate),
                data = cake,
                chains = 1)

bayestestR::describe_posterior(m1, effects = "all")
#> ...
#> 
#> # Random effects SD/Cor: replicate
#> 
#> Parameter          |   Median |         95% CI |     pd |          ROPE | % in ROPE |  Rhat |    ESS
#> ----------------------------------------------------------------------------------------------------
#> (Intercept)        | 2.90e-03 | [ 0.00,  0.70] |   100% | [-0.82, 0.82] |      100% | 0.999 | 612.00
#> temp ~ (Intercept) | 1.86e-04 | [-0.01,  0.01] | 59.20% | [-0.82, 0.82] |      100% | 0.999 | 395.00
#> temp               | 1.12e-03 | [ 0.00,  0.00] |   100% | [-0.82, 0.82] |      100% | 1.000 | 325.00
#> 
#> ...


posteriors <- as.data.frame(m1)

rho <- with(posteriors, {
  `Sigma[replicate:temp,(Intercept)]` / 
    (sqrt(`Sigma[replicate:temp,temp]`) * sqrt(`Sigma[replicate:(Intercept),(Intercept)]`))
})

# Cor
bayestestR::describe_posterior(rho)
#> Summary of Posterior Distribution
#> 
#> Parameter | Median |        95% CI |     pd |          ROPE | % in ROPE
#> -----------------------------------------------------------------------
#> Posterior |   0.19 | [-0.92, 0.95] | 59.20% | [-0.10, 0.10] |     9.05%


# Slopes
a <- with(posteriors, sqrt(`Sigma[replicate:temp,temp]`))

bayestestR::describe_posterior(a)
#> Summary of Posterior Distribution
#> 
#> Parameter | Median |       95% CI |   pd |          ROPE | % in ROPE
#> --------------------------------------------------------------------
#> Posterior |   0.03 | [0.02, 0.05] | 100% | [-0.10, 0.10] |      100%


# Compare to:
lme4::lmer(angle ~ temp + (temp | replicate),
           data = cake) |> 
  lme4::VarCorr()
#>  Groups    Name        Std.Dev. Corr 
#>  replicate (Intercept) 2.604527      
#>            temp        0.027073 0.134
#>  Residual              4.825927
```

mattansb:
Is this maybe an issue in `insight`?

bwiernik:
What does VarCorr() return for that stan_lmer() model?

strengejacke:
> What does VarCorr() return for that stan_lmer() model?

I think this doesn't matter here, as we call `describe_posterior()` and directly work with the posteriors (see https://github.com/easystats/parameters/pull/720#issuecomment-1143213995).

bwiernik:
What I thinking is that I am pretty sure that the model is specified in terms of variance and covariance--the conversion to SD and Corr only happens when VarCorr () is called or the summary is generated

strengejacke:
Ok, got it. Let me check.

strengejacke:
I just realized that `VarCorr()` generally works, but the related methods to extract CI/SE (`vocv()` from _merDeriv_) do not. So we don't have the associated standard errors to compute confidence intervals, or am I missing something?

strengejacke:
> We convert to SD/Corr manually if there isn't a function to do it in rstanarm. corr = cov / sqrt(var1 * var2) for each draw, then compute intervals from the transformed posterior

Ok, but this is indeed something to be done in _bayestestR_

strengejacke:
I opened an issue (https://github.com/easystats/bayestestR/issues/536). Maybe someone can try to address this? I'm not sure about var1/var2, when we have only one posterior of correlation values (on the "wrong" scale). We also need to grep column names to extract random effects... It's a _slightly_ annoying issue, I must admit. ;-)

bwiernik:
How does VarCorr() line up the appropriate parameters?

strengejacke:
> How does VarCorr() line up the appropriate parameters?

What do you mean? How does it extract all parameters, or sorting?

bwiernik:
Re:

> I'm not sure about var1/var2, when we have only one posterior of correlation values (on the "wrong" scale). We also need to grep column names to extract random effects... It's a slightly annoying issue, I must admit. ;-)

VarCorr() lines up the variance and covariance parameters and uses these to compute correlations. Can we copy that logic?

strengejacke:
@IndrajeetPatil Do you know how I correctly include `https://github.com/glmmTMB/glmmTMB/tree/ci_tweaks/glmmTMB` in the "Remotes" section of the DESCRIPTION file?

IndrajeetPatil:
Sorry, I couldn't figure this one out.

The problem is that you need to specify a _subdirectory_ on a given _branch_.

```r
remotes::install_github("glmmTMB/glmmTMB@ci_tweaks", subdir = "glmmTMB")
```

I am not sure what URL this resolves to, and that's exactly what you will need to include in `Remotes`.

IndrajeetPatil:
Let's see if StackOverflow can help: https://stackoverflow.com/questions/72671996/how-to-include-subdirectory-in-remotes-field-for-r-package-description

IndrajeetPatil:
And, of course, some asshole downvoted my question.

IndrajeetPatil:
That worked!

strengejacke:
Nice, thanks! :-)

PR #724, opened 2022-06-06
Support `car::linearHypothesis`

vincentarelbundock:
This function returns an object of class `.anova` with special attributes. So I added a conditional block to process this object in `.extract_anova_anova`. It checks for a special header that should appear in no other anova models, son in principle it shouldn't affect anything else.

The anova `standard_error()` method only calls `model_parameters()`, so it works. I noticed there isn't an `ci.anova()` method, but I can't implement one now.

### DIFF
--- a/R/extract_parameters_anova.R
+++ b/R/extract_parameters_anova.R
@@ -69,7 +69,7 @@
 
   # Reorder
   row.names(parameters) <- NULL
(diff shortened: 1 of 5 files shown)


vincentarelbundock:
Some of this PR was adapted from code by @grantmcdermott here: https://github.com/vincentarelbundock/modelsummary/issues/490 I assume he won't mind, but will let him object if he must.

The test failures appear unrelated to the PR. (1) Mac can't download something using Homebrew, (2) rstanarm on old ubuntu releases.

grantmcdermott:
No objection whatsoever. Thanks for adding!

PR #730, opened 2022-06-11
format CI method name when stored as attribute

IndrajeetPatil:
Examples:

``` r
library(BayesFactor)

mod1 <- ttestBF(x = sleep$extra[sleep$group == 1], y = sleep$extra[sleep$group == 2], paired = TRUE)
p_mod1 <- parameters::parameters(mod1)
attributes(p_mod1)$ci_method
#> [1] "ETI"

mod2 <- lm(wt ~ mpg, mtcars)
p_mod2 <- parameters::parameters(mod2)
attributes(p_mod2)$ci_method
#> [1] "Wald"
```

<sup>Created on 2022-06-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/utils_model_parameters.R
+++ b/R/utils_model_parameters.R
@@ -42,8 +42,8 @@
   }
 
   attr(params, "ci") <- ci
-  attr(params, "ci_method") <- ci_method
-  attr(params, "df_method") <- ci_method
+  attr(params, "ci_method") <- .format_ci_method_name(ci_method)
+  attr(params, "df_method") <- .format_ci_method_name(ci_method)
   attr(params, "test_statistic") <- insight::find_statistic(model)
   attr(params, "verbose") <- verbose
   attr(params, "exponentiate") <- exponentiate
@@ -197,7 +197,19 @@
   params
 }
 
-
+.format_ci_method_name <- function(ci_method) {
+  switch(tolower(ci_method),
+    "eti" = ,
+    "hdi" = ,
+    "si" = toupper(ci_method),
+    "bci" = ,
+    "bcai" = "BCa",
+    "satterthwaite" = ,
+    "kenward" = ,
+    "wald" = paste0(toupper(substr(ci_method, 1, 1)), substr(ci_method, 2, nchar(ci_method))),
+    ci_method
+  )
+}
 
 .find_coefficient_type <- function(info, exponentiate, model = NULL) {
   # column name for coefficients
--- a/R/utils_model_parameters.R
+++ b/R/utils_model_parameters.R
@@ -197,16 +197,25 @@
   params
 }
 
+
+#' Format CI method name when stored as an attribute
+#'
+#' @keywords internal
+#' @noRd
 .format_ci_method_name <- function(ci_method) {
   switch(tolower(ci_method),
+    # abbreviations
     "eti" = ,
     "hdi" = ,
     "si" = toupper(ci_method),
-    "bci" = ,
-    "bcai" = "BCa",
+    # named after people
     "satterthwaite" = ,
     "kenward" = ,
     "wald" = paste0(toupper(substr(ci_method, 1, 1)), substr(ci_method, 2, nchar(ci_method))),
+    # special cases
+    "bci" = ,
+    "bcai" = "BCa",
+    # no change otherwise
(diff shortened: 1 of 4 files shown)


IndrajeetPatil:
pkgdown failure is unrelated (see issues)

strengejacke:
Have you checked that it won't break `.print_footer_cimethod()`? In that function, we have several instances where we compare against lower case strings, like `ci_method != "wald"`, which should be fixed to make sure we don't break the footer.

IndrajeetPatil:
AFAIK, it won't break anything because we are using `toupper()`, which means it doesn't matter what `.format_ci_method_name()` returns:

https://github.com/easystats/parameters/blob/fa78af5fa1b23901f52c249ac443da3d458412f1/R/format.R#L592-L610

strengejacke:
There are several instances in that function where we compare against lower case strings. The two `switch()` statements are the only ones where we convert to upper case.

IndrajeetPatil:
I am now just converting the `ci_method` one more time to lower case within the print function so that everything will work as before.

IndrajeetPatil:
Seems to be working as expected?

``` r
library(BayesFactor)
library(parameters)

mod1 <- ttestBF(x = sleep$extra[sleep$group == 1], y = sleep$extra[sleep$group == 2], paired = TRUE)
parameters(mod1)
#> Bayesian t-test
#> 
#> Parameter  | Median |         95% CI |     pd | % in ROPE |              Prior |    BF
#> --------------------------------------------------------------------------------------
#> Difference |  -1.42 | [-2.29, -0.50] | 99.65% |        0% | Cauchy (0 +- 0.71) | 17.26
#> 
#> Uncertainty intervals (equal-tailed) and p-values (two-tailed) computed
#>   using a MCMC distribution approximation.

mod2 <- lm(wt ~ mpg, mtcars)
parameters(mod2)
#> Parameter   | Coefficient |   SE |         95% CI | t(30) |      p
#> ------------------------------------------------------------------
#> (Intercept) |        6.05 | 0.31 | [ 5.42,  6.68] | 19.59 | < .001
#> mpg         |       -0.14 | 0.01 | [-0.17, -0.11] | -9.56 | < .001
#> 
#> Uncertainty intervals (equal-tailed) and p-values (two-tailed) computed
#>   using a Wald t-distribution approximation.
```

<sup>Created on 2022-06-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/utils_model_parameters.R
```diff
@@ -198,6 +198,31 @@
...
+
+  switch(tolower(ci_method),
+    # abbreviations
+    "eti" = ,
+    "hdi" = ,
+    "si" = toupper(ci_method),
+    # named after people
+    "satterthwaite" = ,
+    "kenward" = ,
+    "wald" = paste0(toupper(substr(ci_method, 1, 1)), substr(ci_method, 2, nchar(ci_method))),
```
strengejacke, line 218: we could use `insight::format_capitalize()` here.

strengejacke: (and we could use your code in that function, it's slightly faster, I think)

strengejacke (changes requested):

strengejacke:
``` r
test <- function(x) {
  if (is.factor(x)) {
    x <- as.character(x)
  }
  if (!is.character(x)) {
    if (verbose) {
      warning(format_message("This function only works on factors or character vector."), call. = FALSE)
    }
    return(x)
  }  
  paste0(toupper(substr(x, 1, 1)), substr(x, 2, nchar(x)))
}

microbenchmark::microbenchmark(
  insight::format_capitalize("uppercaseme"),
  test("uppercaseme"),
  times = 1000
)
#> Unit: microseconds
#>                                       expr  min   lq    mean median   uq
#>  insight::format_capitalize("uppercaseme") 10.7 11.2 42.7276   11.6 11.8
#>                        test("uppercaseme")  6.2  6.5 13.0724    6.9  7.1
```

<sup>Created on 2022-06-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

strengejacke (approved):

PR #734, opened 2022-06-18
Lavaan unknown arguments

vincentarelbundock:
Safer handling of unknown arguments when `standardize=TRUE`:

``` r
library(lavaan)
library(parameters)
model <- ' 
  # latent variable definitions
     ind60 =~ x1 + x2 + x3
'
fit <- cfa(model, data = PoliticalDemocracy)
parameters(fit, effect = "all", standardize = TRUE)
#> Error in lavaan::standardizedsolution(model, se = TRUE, level = ci, type = type, : unused argument (effect = "all")
```


mattansb:
Shouldn't that be "effects"? `parameters(fit, effects = "all", standardize = TRUE)`

vincentarelbundock:
Yeah, the only point of this example is that `effects` is not a supported in `lavaan::standardizedsolution()`. I could have used `junk` or `clown` arguments instead.

PR #766, opened 2022-08-31
P_direction_frq

strengejacke:


R/p_direction.r
```diff
@@ -7,10 +7,11 @@ p_direction.lm <- function(x, dof = NULL, ...) {
   if (is.null(dof)) {
     dof <- degrees_of_freedom(x)
   }
-  out <- .data_frame(
-    Parameter = stats$Parameter,
-    pd = stats::pt(abs(stats$Statistic), df = dof, lower.tail = TRUE)
-  )
+  pd <- stats::pt(abs(stats$Statistic), df = dof, lower.tail = TRUE)
+  if (any(pd < 0.5)) {
```
mattansb, line 11: This is not needed - since you're using `abs`, the smallest this can be is 0.5.

mattansb:
Are we introducing Bayesian probabilistic properties into likelihood-based models :(

strengejacke:
This is a very rough draft, to see whether this makes sense at all:

``` r
library(parameters)
library(rstanarm)
#> Loading required package: Rcpp
#> This is rstanarm version 2.21.3
#> - See https://mc-stan.org/rstanarm/articles/priors for changes to default priors!
#> - Default priors may change, so it's safest to specify priors, even if equivalent to the defaults.
#> - For execution on a local, multicore CPU with excess RAM we recommend calling
#>   options(mc.cores = parallel::detectCores())
#> 
#> Attaching package: 'rstanarm'
#> The following object is masked from 'package:parameters':
#> 
#>     compare_models
library(bayestestR)

m1 <- lm(mpg ~ hp + gear + cyl + am + drat, data = mtcars)
m2 <- stan_glm(mpg ~ hp + gear + cyl + am + drat, data = mtcars, refresh = 0)

p_direction.lm(m1)
#> Probability of Direction
#> 
#> Parameter   |     pd
#> --------------------
#> (Intercept) | 99.75%
#> hp          | 98.33%
#> gear        | 50.08%
#> cyl         | 88.41%
#> am          | 96.42%
#> drat        | 70.88%
p_direction(m2)
#> Probability of Direction
#> 
#> Parameter   |     pd
#> --------------------
#> (Intercept) | 99.80%
#> hp          | 98.30%
#> gear        | 50.12%
#> cyl         | 88.48%
#> am          | 95.58%
#> drat        | 71.28%
```

<sup>Created on 2022-08-31 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

@bmwiernik, what do you think?

R/p_direction.r
strengejacke: ah, ok!

bwiernik:
I'm all for making it easier to get and understand 1-tailed p values!

Implementation-wise, would it be easier/better to just use `p_to_pd()` on the already computed p values? That way things like vcov adjustments, etc are automatically ready

strengejacke:
haha, didn't know we had such a function.

strengejacke:
closing in favor of https://github.com/easystats/bayestestR/commit/62a2e4cd8d2aefcdc3ae079a4f6f1e3c496db4bd
However, we cannot test against specific `null`s, only against 0.

bwiernik:
Do we allow non-nil hypotheses for pd in Bayesian/bootstrap models?

Issue #768, opened 2022-09-03
Model-averaged estimates/intervals/distributions


bwiernik:
There's now a convenient package for computing frequentist model-averaged CIs/CDs https://cran.r-project.org/web/packages/MATA/index.html

As well as for Bayesian model-averaging/stacking https://mc-stan.org/loo/reference/loo_model_weights.html

We should have a function to compute average estimates/intervals/densities. It should permit choice of weighting method, interval method, and be passable to {see} for plotting. 

For frequentist averaging, that's usually done based on information criteria, so this would entail a soft dependency of {parameters} on {performance}.

vincentarelbundock:
Do you see us being able to "extract" info from objects produced by these two packages (probably easy), or do you imagine us writing "wrappers" around those two packages (more maintenance)?

bwiernik:
For both, they take lists of models and return results back, so wrapping isn't likely to be much effort. We already wrap around loo for a much of functions in performance, so we can probably create some unified code for easystats generally

mattansb:
For Bayes, I wrote this function that does this exactly a while back (years???)

https://easystats.github.io/bayestestR/reference/weighted_posteriors.html

Currently used BFs for weights (personally, that's what makes the most sense to me as we are mixing the posteriors from the same data, but different _priors_), but should be expandable (:

bwiernik:
There's a Vehtari, Gelman, et al paper discussing that the "stacking" method used in {loo} by default apparently has some important advantages over traditional Bayesian model averaging

strengejacke:
I think it should be `theta.hats`, right?

MrJerryTAO:
I have used {glmulti} for multimodel inference. I believe it implements frequentist model average by default (and by design???). It explores through millions of possible model formula specifications and retains the top ones of an arbitrary size, optionally using a genetic algorithm, which I have not seen provided in other packages. It also provides model-averaged model summary statistics including coef, SEs, CIs. MuMIn Package was mentioned along with {glmulti} in https://www.metafor-project.org/doku.php/tips:model_selection_with_glmulti_and_mumin and https://www.r-bloggers.com/2013/02/model-selection-and-multi-model-inference, but I have not tried it. 

> Do you see us being able to "extract" info from objects produced by these two packages (probably easy), or do you imagine us writing "wrappers" around those two packages (more maintenance)?

[...]

PR #771, opened 2022-09-05
Model-averaged estimates/intervals/distributions

strengejacke:
cf #768


strengejacke:
``` r
library(MATA)
library(parameters)

set.seed(0)
n = 20                              # 'n' is assumed to be even
x1 = c(rep(0,n/2), rep(1,n/2))      # two groups: x1=0, and x1=1
x2 = rnorm(n, mean=10, sd=3)
y = rnorm(n, mean = 3*x1 + 0.1*x2)  # data generation

x1 = factor(x1)
m1 = glm(y ~ x1)                    # using 'glm' provides AIC values.
m2 = glm(y ~ x1 + x2)               # using 'lm' doesn't.
aic = c(m1$aic, m2$aic)
delta.aic = aic - min(aic)
model.weights = exp(-0.5*delta.aic) / sum(exp(-0.5*delta.aic))
residual.dfs = c(m1$df.residual, m2$df.residual)

p1 = predict(m1, se = TRUE, newdata = list(x1 = factor(0), x2 = 10.64), type = "link")
p2 = predict(m2, se = TRUE, newdata = list(x1 = factor(0), x2 = 10.64), type = "link")
theta.hats = c(p1$fit, p2$fit)
se.theta.hats = c(p1$se.fit, p2$se.fit)

#  95% MATA-Wald confidence interval for theta:
mata.wald(theta.hats, se.theta.hats, model.weights, mata.t = TRUE, residual.dfs = residual.dfs)
#> [1] 0.6196654 1.6516695

parameters:::averaged_parameters(m1, m2)
#> [1] 0.619622 1.651641
```

<sup>Created on 2022-09-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

strengejacke:
@bwiernik implementation seems to work, but I'm not sure about the meaning of this function. It returns _one_ estimate (in the above example currently only CIs)? What does this estimate represent?

bwiernik:
No, that's not right. It should be output with the same structure as the input vectors, averaged across models (eg, if 32 predictions from 2 models, then output should be 32 predictions, weighted across the two models). The MATA CIs for the 32 predictions would then be the values for each prediction that satisfy the critical value when averaged across models

PR #773, opened 2022-09-18
change `data_set` attribute name for `dataset` (in PCA/EFA)

DominiqueMakowski:
It's just nicer to have it in one word, but is it worth it?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Breaking
 
+* The attribute name in PCA / EFA has been changed from `data_set` to `dataset`.
+
 * The minimum needed R version has been bumped to `3.6`.
 
 * Removed deprecated argument `parameters` from `model_parameters()`.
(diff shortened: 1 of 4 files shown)


bwiernik:
The spelling is properly also "dataset" not "data set", so I think the change is good. Changing hidden attributes like this is fine by me

bwiernik (approved):

bwiernik:
What's going on with the tests?

strengejacke:
I think the file [principal_components.R](https://github.com/easystats/parameters/pull/773/files#diff-265b13b9b35251cccfe6c856f7e36cb243be31ef4021671d2a0863234c879937) has more occurrences of `data_set` that need to be changed.

strengejacke:
Note that this is a breaking change and will cause failures in _performance_. Not sure about other packages.

strengejacke (approved):

PR #787, opened 2022-10-02
Remove both rows and columns with complete `NA`s

IndrajeetPatil:
And styling

### DIFF
--- a/R/methods_BayesFactor.R
+++ b/R/methods_BayesFactor.R
@@ -188,9 +188,9 @@ model_parameters.BFBayesFactor <- function(model,
   if (!is.null(out$Effects) && insight::n_unique(out$Effects) == 1) out$Effects <- NULL
 
 
-  # ==== remove columns with complete NA
+  # ==== remove rows and columns with complete `NA`s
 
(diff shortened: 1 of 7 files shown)


R/methods_BayesFactor.R
```diff
@@ -188,9 +188,9 @@ model_parameters.BFBayesFactor <- function(model,
   if (!is.null(out$Effects) && insight::n_unique(out$Effects) == 1) out$Effects <- NULL
 
 
-  # ==== remove columns with complete NA
+  # ==== remove rows and columns with complete `NA`s
 
-  out <- datawizard::remove_empty_columns(out)
+  out <- datawizard::remove_empty(out)
```
IndrajeetPatil, line 193: This is the key change. @strengejacke Do you foresee any issues that this might create?

strengejacke: No, I don't think so. I'm not sure if there's even a situation where we have empty rows.

IndrajeetPatil: Okay, in that case, this shouldn't be problematic at all and may even protect in some weird edge case where all entries in row are missing for some reason.

PR #791, opened 2022-10-05
prepare CRAN submission

strengejacke:


IndrajeetPatil:
Shouldn't _effectsize_ be updated before? 

Because _parameters_ relies on the new API for _effectsize_ updates? 

At any rate, I need to update _statsExpressions_ if _parameters_ is being updated and I can't do so until _effectsize_ is also updated.

@mattansb will you be able to do this anytime soon?

IndrajeetPatil:
> In any case, if you want I can submit.

I'd prefer that, yes.

@strengejacke Can we please also update _insight_? It has been over 2 weeks since the last update, so it should be fine.

strengejacke:
_parameters_ update is on CRAN, and it only breaks _cvms_, so there's no hurry to do anything right now from our side.

IndrajeetPatil:
> there's no hurry to do anything right now from our side

_statsExpressions_ won't break this time, but will fail silently 🙃

strengejacke:
> statsExpressions won't break this time, but will fail silently

Hm, why? I don't think we have breaking changes that affect _statsExpressions_?

strengejacke:
Yes, but I kept the old once, too, for soft-deprecation of this change. :-)

IndrajeetPatil:
Ah, okay. Then we don't need to update immediately. Thanks!

strengejacke:
See, e.g., methods_aov.R:

```r
  ## TODO: remove in a later update
  # handle deprected arguments ------
  if (!is.null(omega_squared)) effectsize_type <- "omega"
  if (!is.null(eta_squared)) effectsize_type <- "eta"
  if (!is.null(epsilon_squared)) effectsize_type <- "epsilon"
```

PR #792, opened 2022-10-05
Remove arguments that have been deprecated for longer now

strengejacke:


bwiernik:
Maybe we should keep the check for df_method and make it an error? To avoid silently returning a result the user isn't expecting

bwiernik:
If we didn't have the dots, including the df_method argument would give an unused argument error. The dots squash that. It think it's a good idea to keep the error because including df_method signals user intent that would otherwise be silently ignored. 

(This is like some of the checks from the ellipsis package)

PR #807, opened 2022-10-31
Export all S3 methods for model_parameters generic - 2nd attempt

jimrothstein:
@IndrajeetPatil 
This branch (jim2) is corrected version.
Checks that each method is actually an S3 generic.

And limits candidates to broom::tidy.X   where the `.` is literal.

broom does have methods of this form, which ARE NOT S3 methods.
broom::tidy_X


jimrothstein:
@IndrajeetPatil 

roxygen2 Embed R code chunks:
(SEE:  CODE CHUNKS:   https://roxygen2.r-lib.org/articles/rd-formatting.html)

Been experimenting.  This code can be put directly into `R/1_model_parameters.R` and will generate all the S3 exports when the command `document()` is run (ie to generate .Rd files and NAMESPACE).   All the manual (cut+paste) @exports for missing S3 methods won't be needed and can be removed.

I've gotten a `warnings` so it is not final shape.
Worth pursing? 

```

#' ```{r test}
#'
#' ## ----------- for parameters
#' ##
#' #	get exported methods tidy.X
#' broom_tidy <- ls(getNamespace("broom"), pattern="^tidy[.]")
#' 
#' 
#' #	check every candidate method is one of  method(tidy)
#' broom_tidy %in% methods(broom::tidy) 
#' if (!all(broom_tidy %in% methods(tidy) )) stop("Check all functions are actual S3 methods")
#' 
#' 
#' #	get classes
#' broom_supported <- gsub("^tidy[.]", "", broom_tidy)
#' 
#' 
#' 
#' #	get exported methods model_parameters.X
#' easystats_tidy <- ls(getNamespace("parameters"), pattern="^model_parameters[.]")
#' easystats_tidy
#' 
#' 
#' #	check every candidate method is one of  method(model_parameters)
#' if (!all(easystats_tidy %in% methods("model_parameters") )) stop("Check all functions are actual S3 methods")
#' 
#' 
#' # get classes
#' easystats_supported <- gsub("^model_parameters[.]", "", easystats_tidy)
#' 
#' 
#' #	helper function , list every elemnt of x that is NOT in y
#' "%w/o%" <- function(x, y) x[!x %in% y] #--  x without y
#' 
#' missing  <- broom_supported %w/o% easystats_supported
#' 
#' cat(vapply(missing, 
#' 	function(x) { paste0("#' @export\n", "model_parameters.",x, "  <- model_parameters.default\n\n")},
#' 	  character(length = 1), 
#' 		USE.NAMES=F)
#' 
#' )
#' 
#' ```
```

strengejacke:
I'm not sure, but I think this can be merged, no?

jimrothstein:
@strengejacke 
Unless you see another error, I think it should be merged.  It includes check that `methods` are true S3 methods.

IndrajeetPatil:
@strengejacke There seems to be a test failing, but I am not sure if it is related to this PR. My guess is it isn't. But the `main`-branch is failing as well, so no way to tell.

```r
== Failed tests ================================================================
-- Error ('test-model_parameters.aov_es_ci.R:262'): works with manova ----------
Error: Error: Sorry, `model_parameters()` failed with the following error (possible
  class `manova` not supported):

  arguments imply differing number of rows: 0, 1
Backtrace:
    x
 1. +-base::as.data.frame(...) at test-model_parameters.aov_es_ci.R:262:6
 2. +-parameters::model_parameters(...)
 3. \-parameters:::model_parameters.manova(...) at parameters/R/1_model_parameters.R:291:2
 4.   \-insight::format_error(...) at parameters/R/1_model_parameters.R:505:4
 5.     \-insight::format_alert(..., type = "error")

[ FAIL 1 | WARN 1 | SKIP 7 | PASS 1226 ]
```

strengejacke:
oh, right. `manova` also inherits from `aov`, so 

```r
 #' @export
model_parameters.manova  <- model_parameters.default
```

should be

```
 #' @export
model_parameters.manova  <- model_parameters.aov
```

This makes it more complicated than expected, because we have to check which models inherit from multiple classes, and those probably should not be redirected to the `.default` method.

strengejacke:
Looks like the failing checks are due to some config-errors or temporary?

IndrajeetPatil:
Yeah, seems like some issue in `{pak}` or something related. Will have a closer look later.

jimrothstein:
@IndrajeetPatil @strengejacke 

For time being, should make this change, and resubmit?

` #' @export
model_parameters.manova  <- model_parameters.default

should be

 #' @export
model_parameters.manova  <- model_parameters.aov
`

strengejacke:
Yes. Thanks!

jimrothstein:
@strengejacke      What can or should I be doing at this point?  (Working through vignettes, but hardly up-to-speed yet)

IndrajeetPatil:
@strengejacke Shouldn't we merge this (after resolving conflicts)?

strengejacke:
I'm a bit hesitant, unless we have carefully checked each method. E.g., we now have `model_parameters.geeglm`, which runs `model_parameters.default`. However, geeglm inherits also from `glm`, so the former behaviour was that geeglm models used `model_parameters.glm` instead of the default method. This can result in unexpected behaviour.

That means we can only safely refer to the default methods for those models that only inherit from one single class for which we have no method yet. And this needs to be checked.

``` r
library(geepack)

data(dietox)
dietox$Cu     <- as.factor(dietox$Cu)
mf <- formula(Weight ~ Cu * (Time + I(Time^2) + I(Time^3)))
gee1 <- geeglm(mf, data=dietox, id=Pig, family=poisson("identity"), corstr="ar1")
class(gee1)
#> [1] "geeglm" "gee"    "glm"    "lm"
```

PR #812, opened 2022-11-20
Be more flexible for HTML print

strengejacke:


strengejacke:
image

Only thing missing is how to remove the duplicated column header label from the sub columns. But `gt` requires unique column names...

strengejacke:
image

IndrajeetPatil:
Yes, these failures are expected. 

A few of them are due to the fact that interaction symbol has changed since the snapshots were created (`×` vs `*`).

You can check what has changed in outputs using `testthat::snapshot_review()`. Or just run `testthat::snapshot_accept()` and then study the changes before committing them.

strengejacke:
I already accepted all tests... So maybe I just change the symbol to `*` then.

Issue #806, opened 2022-10-27
Missing `pretty_labels` for variance components


vincentarelbundock:
Should the `pretty_labels` attribute include labels for all elements of the `parameters()` output? I would personally have expected them there...

``` r
library(lme4)
#> Loading required package: Matrix
library(parameters)
data(Orthodont, package = "nlme")
Orthodont$nsex <- as.numeric(Orthodont$Sex == "Male")
m <- lmer(distance ~ age * nsex + (1 | Subject), data = Orthodont)

p <- parameters::parameters(m)
data.frame(p)
#>           Parameter Coefficient         SE   CI      CI_low    CI_high
#> 1       (Intercept)  17.3727273 1.18350709 0.95 15.02524658 19.7202080
#> 2               age   0.4795455 0.09346985 0.95  0.29414845  0.6649425
#> 3              nsex  -1.0321023 1.53742081 0.95 -4.08156915  2.0173646
#> 4          age:nsex   0.3048295 0.12142090 0.95  0.06399177  0.5456673
#> 5    SD (Intercept)   1.8162142 0.29501895 0.95  1.32099484  2.4970833
#> 6 SD (Observations)   1.3863819 0.11029462 0.95  1.18621913  1.6203202

attr(p, "pretty_labels")
#>   (Intercept)           age          nsex      age:nsex 
#> "(Intercept)"         "age"        "nsex"  "age * nsex"
```

Above, we are missing pretty labels for the variance components.

strengejacke:
We simply could add the missing Parameter string values to the `pretty_labels` resp. `pretty_names` attribute.

strengejacke:
(and hope that it doesn't break anything)

PR #816, opened 2022-11-26
Issue #806

vincentarelbundock:
#806


strengejacke:
Regarding #891, we also need to take care of code like https://github.com/easystats/parameters/blob/59a569c8d3d9997996145d9e16d0e9aeea7a83c8/R/format_parameters.R#L376

While the model frame has the response variable in the first column, this may be no longer true when we recover data from the environment. We should retrieve the response column name using `find_response()` and `setdiff()` it from the column names?

PR #822, opened 2022-12-14
Simplify conditional expressions in tests

IndrajeetPatil:
cf. https://github.com/easystats/easystats/issues/310


vincentarelbundock:
@IndrajeetPatil I think the `if(requiet(.....))` are not even necessary because `requiet` includes a `skip_*` call. So in principle, we could get rid of the conditionals altogether.

IndrajeetPatil:
Yeah, but this is hard to do by search-and-replace. So I am keeping that cleanup for a separate PR.

PR #826, opened 2022-12-22
check on linux, to increase codecov

strengejacke:


strengejacke:
Ok, can someone tell me why we have this test failure on one Linux system, but not the other?

```
-- Failure ('test-simulate_model.R:51'): simulate_model, glmmTMB ---------------
head(s$child) (`actual`) not equal to c(-1.21946, -1.23724, -1.10968, -1.14867, -1.04882, -1.11192) (`expected`).

  `actual`: -1.322 -1.200 -1.145 -1.199 -1.186 -1.118
`expected`: -1.219 -1.237 -1.110 -1.149 -1.049 -1.112
-- Failure ('test-simulate_model.R:64'): simulate_model, glmmTMB, conditional only --
head(s$child) (`actual`) not equal to c(-1.21946, -1.23724, -1.10968, -1.14867, -1.04882, -1.11192) (`expected`).

  `actual`: -1.322 -1.200 -1.145 -1.199 -1.186 -1.118
`expected`: -1.219 -1.237 -1.110 -1.149 -1.049 -1.112
```

IndrajeetPatil:
Yeah, I am stumped by this as well.

I'd have expected it to fail for all R versions of Ubuntu.

IndrajeetPatil:
The only (temporary?) solution I can think of is to choose a liberal threshold and add a `TODO` to revisit this later?

vincentarelbundock:
Which Linux/R versions is this? Before the big RNG change?

strengejacke:
The failing test is on

```
- { os: ubuntu-latest, r: "devel", locale: "en_US" }
```

Interestingly, the same test does not fail on a different locale:
```
- { os: ubuntu-latest, r: "devel", locale: "zh_CN" }
```

@IndrajeetPatil Any ideas where these systems differ, except their locale?

etiennebacher:
I can explore this later on my Ubuntu machine

etiennebacher:
> I can explore this later on my Ubuntu machine

I can't reproduce the errors on Ubuntu with R-devel, I tried with both locales and both work well. I didn't try with oldrel-3 (which R release is it?) but the error is different, maybe there's just a package in Suggests that can't be installed with this R version?

IndrajeetPatil:
I think this can be closed since we now compute code coverage on Windows machines.

Issue #849, opened 2023-02-27
Revise the way we calculate % in ROPE (SGPV) for frequentist models?


strengejacke:
This idea came up after a chat with Isabella on Twitter.

See following example at https://easystats.github.io/parameters/reference/equivalence_test.lm.html

``` r
library(parameters)
data(qol_cancer)
model <- lm(QoL ~ time + age + education, data = qol_cancer)
equivalence_test(model)
#> # TOST-test for Practical Equivalence
#> 
#>   ROPE: [-1.99 1.99]
#> 
#> Parameter        |         90% CI | % in ROPE |        H0 |      p
#> ------------------------------------------------------------------
#> (Intercept)      | [59.33, 68.41] |        0% |  Rejected | > .999
#> time             | [-0.76,  2.53] |    83.52% | Undecided | 0.137 
#> age              | [-0.26,  0.32] |      100% |  Accepted | < .001
#> education [mid]  | [ 5.13, 12.39] |        0% |  Rejected | 0.999 
#> education [high] | [10.14, 18.57] |        0% |  Rejected | > .999
```

If we look at the parameter `time`, we see a CI of `(-0.76, 2.53)` and 83.52% inside ROPE. The latter is equivalent to the _SGPV_. Both SGPV and % in ROPE are calculated the same way (see GitHub repo from Daniel: https://github.com/Lakens/TOST_vs_SGPV/tree/master/functions, related paper https://open.lnu.se/index.php/metapsychology/article/view/933).

However, although we actually want the overlap of the _interval_ (probability mass) with the ROPE, it's calculated (for simplicity) based on the overlap of the _range_.

What we are **actually doing** is:

``` r
# hypothetical min/max
est.hi <- 2.53
est.lo <- -0.76

x <- bayestestR::distribution_uniform(
  n = 500,
  min = est.lo,
  max = est.hi
)
# actual (simulated) min/max
min(x)
#> [1] -0.75671
max(x)
#> [1] 2.52671

bayestestR::equivalence_test(x, c(-2, 2), ci = 1)
#> # Test for Practical Equivalence
#> 
#>   ROPE: [-2.00 2.00]
#> 
#> H0        | inside ROPE |     100% HDI
#> --------------------------------------
#> Undecided |     83.80 % | [-0.76 2.53]
bayestestR::rope(x, c(-2, 2), ci = 1) |> plot()
```

![](https://i.imgur.com/iqi6l4C.png)

But what we **actually want**, is:

``` r
[...]

PR #850, opened 2023-02-27
Revise the way we calculate % in ROPE (SGPV) for frequentist models? #849

strengejacke:
#849


bwiernik:
My preference for both freq and Bayesian would be to default to computing as the proportion of the posterior that's in the ROPE (so `rope_ci = 1`). That would make the % be 1 - the equivalence test p value IIRC

strengejacke:
I think this doesn't work right now. We have this code, with hardcoded `3.28`:

```r
  x <- bayestestR::distribution_normal(
    n = 1000,
    mean = ci_range[2] - (diff_ci / 2),
    sd = diff_ci / 3.28
  )
```

I guess this works approximately only for 95% CI, not for others...

PR #856, opened 2023-03-14
alert not warn

strengejacke:


vincentarelbundock:
Out of curiosity, do you think I should expect a lot of printouts to the console in modelsummary or will this only affect special cases -- like bad user input -- that actually need attention from the user?

strengejacke:
> Out of curiosity, do you think I should expect a lot of printouts to the console in modelsummary

The only thing I changed is that some `warnings()` are now `messages()`. There are no new alerts, and `verbose` still suppresses them as before. Unless you used `suppressWarnings()` to suppress output, nothing should change. Else, you may need to use `suppressMessages()` instead.

But in general, the next release round (https://github.com/easystats/easystats/issues/355) should even better deal with `verbose`, so that messages/warnings that were still printed even with `verbose = FALSE`, should now be suppressed.

strengejacke:
I know @bwiernik doesn't like warnings, and I revisited some of the warnings and indeed, maybe a message is enough in those cases I changed in this PR. It's mostly an information, nothing where a warning should indicate serious flaws.

PR #860, opened 2023-03-23
test w/o remotes

strengejacke:


strengejacke:
@IndrajeetPatil I'm not sure whether [this issue](https://github.com/easystats/parameters/actions/runs/4499548428/jobs/7917553814?pr=860) is something to be fixed in _see_ (I think so), do you know if this is resolved?

IndrajeetPatil:
Yes, I think this is already fixed in the devel-_see_ but will double check.

PR #862, opened 2023-03-23
Remove data frame methods

IndrajeetPatil:
I checked both _parameters_ and _bayestestR_ with these changes, and none of the tests seem to fail.


NEWS.md
```diff
@@ -2,9 +2,13 @@
...
 ## General
 
-* Added support for models of class `hglm` (*hglm*), `mblogit` (*mclogit*) and
+* Added support for models of class `hglm` (*hglm*), `mblogit` (*mclogit*), and
   `fixest_multi` (*fixest*).
 
+* `as.data.frame` methods for regression model objects have been removed. The
+  `model_parameters()` function is the preferred solution to tidy regression
+  model objects into a data frame.
+
```
strengejacke, line 11: I think their intention was not to be an alias for `model_parameters()`, but rather to behave like `as.data.frame.brmsfit()` and return posterior draws via `bootstrap_model()`.

strengejacke (changes requested):

NEWS.md
IndrajeetPatil: But where is this used?

IndrajeetPatil: Or the change you are requesting is only editing the NEWS item?

strengejacke: > I don't think that the as.data.frame() methods for bootstrapping should be retired. They provide useful parity between Stan models and other models

I agree they're convenient, but the problem here is that we're introducing a breaking change into the behaviour of `as.data.frame()` for base-R objects. This _can_ lead to unintended behaviour and non-working code that is hard to debug then. We had this with `datawizard::data_read()`, which checks whether an object can be coerced to data frame, and if not, returns the object itself (which is sometimes intended, when you have an RDS file with a model object, e.g.). The tests worked well and everything passed, but then one test was run before, which loaded the _parameters_ package. The RDS file, containing an `lm` object, now suddenly returned a data frame for `data_read()`, which led to weird results and we first couldn't figure out why.

Since the `as.data.frame()` methods really just call `bootstrap_model()`, there's not much lost to retire them.

IndrajeetPatil: I think the convenience afforded by these methods is more than offset by the potential problems they can cause, and that too in an unpredictable and mysterious ways, as we painfully discovered in the case of _datawizard_.

IndrajeetPatil:
@strengejacke You don't need to keep rebasing PR with the main for every new commit. This is wasteful and also sends out a new email to all PR contributors on every single commit. 

Just do it once when the PR is ready for a merge, assuming, of course, that the changes you are pushing to the main are in a completely different parts of the codebase than the one in the PR.

strengejacke:
> sends out a new email
I never receive any mail from GH?

rempsyc:
> I never receive any mail from GH?

I had to deactivate notification parameters for this, was on per default

notifications

PR #869, opened 2023-04-03
fix vignettes

strengejacke:


IndrajeetPatil:
Maybe use the idiom mentioned here in _parameters_ as well?

https://github.com/easystats/easystats/issues/341

PR #903, opened 2023-09-13
CRAN submission 0.21.2

strengejacke:
- addresses https://github.com/easystats/easystats/issues/318


strengejacke:
@IndrajeetPatil I think I already asked this, but I forgot how to address this:
check-strict fails on Ubuntu for this reason:

> ❯ checking package dependencies ... NOTE
>  Packages suggested but not available for checking: 'cplm', 'panelr'

Is there any way to avoid that this note results in a check failure?

strengejacke:
And this issue: https://github.com/easystats/parameters/actions/runs/6174249881/job/16758433068?pr=903

Is it temporary? I can't exactly figure out what's going on here.

IndrajeetPatil:
> And this issue: https://github.com/easystats/parameters/actions/runs/6174249881/job/16758433068?pr=903
> 
> Is it temporary? I can't exactly figure out what's going on here.

This should be fixed. Dependency graphs have changed and so some more packages needed to be ignored on older versions.

IndrajeetPatil:
> @IndrajeetPatil I think I already asked this, but I forgot how to address this: check-strict fails on Ubuntu for this reason:
> 
> > ❯ checking package dependencies ... NOTE
> > Packages suggested but not available for checking: 'cplm', 'panelr'
> 
> Is there any way to avoid that this note results in a check failure?

I had already set the needed [environment variable to false](https://github.com/easystats/workflows/blob/374eef1d51f7435cde0477ccbd1bbeb6b6423792/.github/workflows/R-CMD-check-strict.yaml#L32), so I honestly don't understand why the R CMD check is still failing.

PR #904, opened 2023-09-16
PCA with n = 1

strengejacke:


R/principal_components.R
```diff
@@ -419,14 +419,8 @@ principal_components.data.frame <- function(x,
     n <- ncol(x) - 1
   } else if (n >= ncol(x)) {
     n <- ncol(x) - 1
```
bwiernik, line 421: I would say this should actually be setting `n <- ncol(x)`. Plausibly someone could want to just transform variables into principal components without any dimension reduction 

@strengejacke

PR #948, opened 2024-02-11
parameters.marginaleffects - preserve attributes from original input

DominiqueMakowski:


DominiqueMakowski:
Doesn't work:

``` r
  model <- lm(Sepal.Width ~ Species * Petal.Length, data = iris)
  
  means <- marginaleffects::predictions(model, newdata=insight::get_data(model), by="Species")
  attr(means, "SOMETHING") <- "Yes"
  attributes(parameters::parameters(means))$SOMETHING 
#> NULL
```

<sup>Created on 2024-02-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

DominiqueMakowski:
and I can't seem to trace where it gets overwritten again 🤔

vincentarelbundock:
As far as I can tell, `parameters` never preserves the attributes of the model object. Try 

```r
mod <- lm(mpg ~ wt * factor(cyl), data = dat)
attr(mod, "SOMETHING") <- "yes"
attributes(parameters::parameters(mod))$SOMETHING 
```

DominiqueMakowski:
Mmh fair. I don't *need* this, it was mostly for convenience reasons in modelbased but I can do without

PR #998, opened 2024-07-23
remove degrees_of_freedom methods

strengejacke:


strengejacke:
@vincentarelbundock I don't know if you call `degrees_of_freedom()` in one of your packages - it's now completely moved to `insight::get_df()`. From a user perspective, nothing should have changed. But since all methods were removed here and `degrees_of_freedom()` just calls `insight::get_df()` now, there may be new errors not detected yet.

My question: once this PR is merged, could you test your packages (I'm particularly thinking of _modelsummary_) with the GitHub version of _parameters_?

vincentarelbundock:
Thanks for the heads-up. I think I had already moved everything to `get_df()`, so this should be no problem. I ran the test suite locally with Github versions of all `easystats` packages I depend on, and didn't seen any error.

Consolidation is a good move here. Well done!

PR #1030, opened 2024-10-17
Prepare CRAN submission

strengejacke:


DominiqueMakowski:
```
Examples with CPU (user + system) or elapsed time > 10s
                    user system elapsed
estimate_contrasts 11.19   0.81   12.03
```

It seems like it's always these examples, we should probably deactivate most of them on CRAN and rely on GH tests

strengejacke:
You're reporting the win-builder checks for modelbased?

strengejacke:
Moving your comment to https://github.com/easystats/modelbased/pull/266

PR #1070, opened 2025-03-04
Prepare CRAN release

strengejacke:


strengejacke:
@DominiqueMakowski For now, I renamed that option in `"random_total"`, but I kept the alias `"total"` (not exposed), for compatibility to _modelbased_.

PR #1080, opened 2025-04-15
Update article current links

guspan-tanadi:
past [archived](https://web.archive.org/web/20200704215418/http://towardsdatascience.com/stopping-stepwise-why-stepwise-selection-is-bad-and-what-you-should-use-instead-90818b3f52df) version of stepwise selection.


strengejacke:
Thanks, good catch!

PR #1122, opened 2025-06-28
Add test for `reverse_items`

strengejacke:


strengejacke:
> People usually later still need to compute averages / sums of dimensions so they will need to reverse the items explicitly, so they might as well do it before, especially since we have a nice datawizard::reverse_scale() function

We also have this item in `get_scores()`, and if you do it once in `factor_analysis()` or `principal_components()`, the "reversed" data is stored as attributes, for further computation.

But I have no string opinion here, we can have both. One argument _against_ having a `reverse_item` argument in that people have to remember in subsequent processing whether they're are working with the reversed data set or not, which is not that explicit when you

1. input "not reversed" data to `factor_analysis()`
2. `factor_analysis()` stores the "reversed" data as attribute
3. Further functions like `get_scores()` or `check_itemscale()` applied on the object returned by `factor_analysis()` use this already "reversed" data, so no need to use the `reverse_items` argument in those post-processing functions again (if you do it, you double reversed, meaning you have the "unreversed" data again)

PR #1209, opened 2026-04-26
typos and wordsmithing on 'demean' vignette

bbolker:
* Some straightforward typographical corrections, although a few of them (substitution of em-dashes for hyphens where appropriate, a little bit of wordsmithing) are a little more opinionated
* I changed the plot for "Model 2" (fixed effect) to use the correct model, i.e. to use a FE model to get parallel slopes, which is not what happens when using `geom_smooth(method = "lm")`)
* I think Model 4 should also be changed to correspond to the model actually fitted (which may look *very* slightly different from Model 2 as there is also shrinkage in the intercept estimates). I did the same for Model 4 (although I'm not sure that the predictions are actually any different, since the fit is singular?  Is this something you want to mention in a footnote?)
* Is it worth adding a figure for Model 5 that would show the predictions of the random-slopes model? I can do that if you like ...
*

### DIFF
--- a/vignettes/demean.Rmd
+++ b/vignettes/demean.Rmd
@@ -413,11 +413,15 @@ and include only the within-effects as well as the group-level indicator.
 
 ```{r fe-model-preds, echo = FALSE}
 fe_model <- lm(y ~ 1 + grp + x, data = d)
-fe_pred <- d |>
-  select(-y) |>
-  reframe(x = range(x), x_within = range(x_within),
-          x_between = x_between[1], .by = grp)
-fe_pred$y <- predict(object = fe_model, newdata = fe_pred)
+sumfun <- function(df) {
+  data.frame(
+    grp = df$grp[1],
+    x = range(df$x),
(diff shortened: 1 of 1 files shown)


strengejacke:
Thanks a lot! The `::` is rather for R code files, not rmd files, but the gemini.md instructions are probably not precise enough here.

> I did the same for Model 4 (although I'm not sure that the predictions are actually any different, since the fit is singular? Is this something you want to mention in a footnote?)

Not sure, I see vignettes mainly to demonstrate how to do something, not necessarily to deal with model fitting issues - unless those issues are in the focus of the vignette. But we can add a footnote, anyway, just in case someone is confused.

> Is it worth adding a figure for Model 5 that would show the predictions of the random-slopes model? I can do that if you like ...

Sure, that would be great!

bbolker:
Done. (If I went into too much detail in the explanation, feel free to cut it back.)

strengejacke:
Great, thanks a lot!

PR #1219, opened 2026-05-15
BUG FIX: bootstrap_model() ignored argument 'cluster'

HenrikBengtsson:


strengejacke:
Thanks! I applied the air-formatting of the file, which wasn't done yet (because we didn't touch that file after we used air to format the code). That's why there are so many "changes" now.

Will run checks now.

Issue #1110, opened 2025-06-05
`tinyplot` method(s)?


grantmcdermott:
This might be encroaching on `see`'s toes, but is there any possibility of supporting [tinyplot](https://grantmcdermott.com/tinyplot/) methods for `parameters`? It would only require a lightweight `Suggests` addition and I think we could keep the implementation pretty simple. Something along the lines of:

``` r
tinyplot.parameters_model = function(
  x,
  flip = TRUE,
  zero = TRUE,
  ...
) {
  xlabs = attr(x, "pretty_labels")
  x$Parameter = xlabs[x$Parameter]
  tinyplot(
    Coefficient ~ Parameter, ymin = CI_low, ymax = CI_high,
    data = x,
    type = "pointrange",
    flip = flip,
    ...
  )
  if (zero) if (flip) abline(v = 0, lty = 2) else abline(h = 0, lty = 2)
}
```

MWE:

``` r
library(tinyplot)
library(parameters)

res = parameters(lm(wt ~ am * cyl, data = mtcars))
plt(res)
```

![](https://i.imgur.com/xhNXsIl.png)

Slightly fancier version with a theme:

``` r
tinytheme("classic")
plt(res)
```

![](https://i.imgur.com/OYYhC4p.png)

<sup>Created on 2025-06-04 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Happy to iterate on a potential PR if there's interest on your side, but no offence if it's out of scope.

strengejacke:
Sure! I was indeed thinking already about this, also regarding a broader support.

Maybe even support for [visualization recipes](https://easystats.github.io/datawizard/reference/visualisation_recipe.html). Basically, this sets up a layer with ggplot geoms/aes, which is more or less simply plotted using [`geoms_from_list()`](https://easystats.github.io/see/reference/geom_from_list.html). This approach has the advantage that we can have the plot preparation inside the related packages, without adding dependencies, and the _see_ package is only needed for a simple plot method. But that's probably an overkill for now.

I'm happy to have tinyplot methods for the easystats packages, especially modelbased. Maybe we can even use it as alternative backend for some _performance_ plots.

strengejacke:
For `model_parameters()`, also consider models with multiple components (zero-inflation) or group levels of random effects. The full overview of possible plot-support is in the related vignette from the _see_ package: https://easystats.github.io/see/articles/parameters.html

strengejacke:
[...]

PR #1244, opened 2026-07-23
add tinyplot method for parameters_model objects

haomeng797-ship-it:
Addresses #1110.

Adds a tinyplot() / plt() forest plot method for parameters_model objects, building on @grantmcdermott's sketch in the issue:

- pointrange forest plot with pretty parameter labels, flipped by default, first parameter on top (via the new smart axis limits, xlim = "rev")
- dashed reference line at zero, or at one for exponentiated coefficients; the estimate axis always covers the reference value (ylim = 0 scalar form from tinyplot 0.7.0, as suggested in the thread)
- sort = TRUE orders coefficients largest on top
- mixed models drop random effects, zero-inflation models show the conditional component, both with a message; multinomial response levels are disambiguated in the labels
- axis label follows the coefficient type ("Odds Ratio" etc.)

One note: this method is also needed to intercept dispatch, since tinyplot 0.7.0 added tinyplot.data.frame(), so a bare plt(x) on a parameters_model currently draws a pairs-style grid rather than a coefficient plot.

tinyplot (>= 0.7.0) added to Suggests. Tests cover lm, glm (exponentiated), lmer, zeroinfl, multinom and polr, plus three vdiffr snapshots. Lints, spelling and Rd checks pass locally. Happy to extend to other classes (simulated/Bayesian parameters) in a follow-up if this direction looks right.


haomeng797-ship-it:
Thanks for the review! Addressed both points:

- The component message now names the component that is actually shown (e.g. "Only the count component is plotted.") instead of always saying "conditional".
- xlim = "rev" is now only set when flip = TRUE, so a vertical plot keeps the parameters in model order.

Also ran air, so the formatting check should pass now.

strengejacke:
I'll also do a personal review later, but sometimes the AI review can be helpful spotting things which I miss, so I routinely ask co pilot for a review first

haomeng797-ship-it:
One note: the format-suggest check is red only because pull_request_target refuses to check out fork code, not because of a formatting issue. I ran air format . locally and it comes back clean, so this should sort itself out once it runs from the base. Happy to double-check if you'd like.

strengejacke:
You may add yourself as contributor to the DESCRIPTION file, if you like (unless you prefer to "stay anonymous" 😉)

haomeng797-ship-it:
Thanks, I have added myself as a contributor in the DESCRIPTION.

strengejacke:
Looks good to me, thanks a lot!

PR #1245, opened 2026-08-04
Save the `effects` argument so group-level estimates print as random effects

### LINKED ISSUE #1098: Fix printing for  `effects = "grouplevel"`
We need to save the `effects` argument, so we can detect whether we have `"grouplevel"` in `.format_columns_multiple_components()`.

Currently, group-level estimates have the wrong header, because the `Effects` column is removed, since it only has one unique value (`"random"`). This means that group level effects are not detected as random effects when printing.

haomeng797-ship-it:
Fixes #1098.

With `effects = "grouplevel"` the `Effects` column carries a single unique value, so it is dropped before formatting, and both header paths then fall through to fixed effects. The clearest case is a zero-inflated brms model, where the group-level estimates printed as `# Fixed Effects (persons)` and `# Zero-Inflation Parameters (persons)`; with this PR they print as `# Random Effects: conditional (persons)` and `# Random Effects: Zero-Inflation (persons)`, matching what the equivalent glmmTMB model already did.

As suggested in the issue, `.add_model_parameters_attributes()` now stores the `effects` argument as an attribute. Two details:

- Several methods (glmmTMB, coxme, and the merMod path) resolve the `"grouplevel"` alias into `effects = "random"` plus `group_level = TRUE` before the attributes are added, so the attribute is normalized back to `"grouplevel"` from that pair, and it always records what the user asked for.
- The mixed-capable call sites (default method, brms, glmmTMB, coxme, rstanarm, stanfit) now forward `effects`; for other methods the attribute is `NULL` and nothing changes.

`.format_columns_multiple_components()` and `.print_caption()` consult the attribute as a fallback when the `Effects` column is gone, so lmer keeps its `# Random Effects` caption and the multi-component tables get the `random.` prefix they were missing.

Tests: a new `test-print-grouplevel.R` covers the lmer caption and the brms case (header lines only, no value snapshots, since `brms_zi_3` snapshots were already flagged as unstable in test-printing-stan.R). Existing printing and stan-printing suites pass unchanged, so no snapshots move.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Type: Package
 Package: parameters
 Title: Processing of Model Parameters
-Version: 0.29.2.3
+Version: 0.29.2.4
 Authors@R:
     c(person(given = "Daniel",
              family = "Lüdecke",
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,14 @@
 * Logical variables are now properly formatted when pretty value labels are
   printed.
 
+## Bug fixes
+
+* The `effects` argument is now saved as an attribute of the returned
+  parameters table, and printing uses it to recognize group-level estimates
+  as random effects. Previously, when `effects = "grouplevel"` produced a
+  single-valued `Effects` column that was removed before formatting, the
+  estimates were printed under fixed-effects headers (#1098).
+
 # parameters 0.29.2
 
 ## Changes
--- a/R/1_model_parameters.R
+++ b/R/1_model_parameters.R
@@ -731,6 +731,7 @@ model_parameters.default <- function(
         ci_method = ci_method,
         p_adjust = p_adjust,
         include_info = include_info,
+        effects = effects,
         verbose = verbose,
         ...
       )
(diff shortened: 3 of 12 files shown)


haomeng797-ship-it:
CI note: the R-CMD-check, pkgdown and coverage failures reproduce on current main (a `range` validation error coming from a dependency, plus a lavaan warning), and the styling/spelling flags predate this diff; the spelling hit is the word "errored" in the 0.29.2 section of the changelog. The jobs that gate on the changed code itself (lint, check-all-examples) pass.

strengejacke:
Thanks for this PR! Yes, please ignore the CI checks. The one that breaks packagedown website building is already fixed in the _performance_ package, and the other test with _lavaan_ only works interactively - I'll skip that one.

Let me review this PR after I triggered the Co-pilot AI review.

strengejacke:
Thanks a lot, looks good to me! 
Can you please increase the version number in the DESCRIPTION file (increase fourth digit by 1). You also may add yourself as contributor, if you like (unless you would like to stay incognito).

strengejacke (approved):

haomeng797-ship-it:
Done, version bumped to 0.29.2.4. And thanks for the offer! I am already in the DESCRIPTION from #1244, so nothing to add there.

PR #1253, opened 2026-09-07
Fix standard errors for glmmTMB ordinal() models (needs easystats/insight#1221)

jmgirard:
> [!IMPORTANT]
> Do not merge before easystats/insight#1221. This depends on `insight::get_varcov()` returning the delta-method covariance matrix for glmmTMB's `ordinal()` family, which lands there (hence `insight (>= 1.5.3.2)` in DESCRIPTION). CI will fail until r-universe rebuilds insight.

Companion to easystats/insight#1221 for the new `ordinal()` family in glmmTMB (>= 1.1.15).

`.se_fixed_effects_glmmTMB()` reads standard errors from `coef(summary(model))` and labels them with `find_parameters()`. For ordinal fits the thresholds are not in the summary table (they are family parameters estimated on an internal softmax scale), so the two vectors differ in length: `standard_error()` and `model_parameters()` either errored, or worse, silently recycled the slope SEs over the thresholds (e.g. SE of `1|2` reported as 0.60 instead of 0.68 on `ordinal::wine`).

This PR takes the SEs for ordinal fits from `insight::get_varcov()` instead, which returns them on the threshold scale via the delta method and aligned by parameter name (non-estimated parameters get `NA`). `ci()` and `p_value()` were already routed through insight and unaffected.

Verified against `ordinal::clm()`/`clmm()`: `model_parameters()` now matches column for column (Coefficient, SE, CI, z, p), including probit link, `exponentiate`, `ci_method = "profile"`, `bootstrap = TRUE`, `standardize = "refit"`, rank-deficient fits, and `compare_parameters()`. Tests added in `test-glmmTMB-ordinal.R` (skipped unless glmmTMB >= 1.1.15 and insight >= 1.5.3.2). Full test suite and `R CMD check` clean against the dev insight.


strengejacke:
Thanks, looks good to me! Will run local checks.

strengejacke:
@jmgirard Is emmeans-support already implemented in glmmTMB?

strengejacke:
I think the copilot comments are either false-positives or not critical. Merging this PR.

